PR #4, opened 2025-11-05
spec: AI card and AI catalog draft proposal

mindpower:
## Initial Draft of AI Card & AI Catalog Specifications

 *Note: This PR has been updated to reflect the converged schema architecture co-authored and designed in collaboration with @muscariello .*
 **Key Architecture Decisions:**
 * **Separation of Concerns:** Core identity and trust remain at the root.
 * **Isolated Namespaces:** Live API interactions are routed through the `protocols` map (e.g., `protocols.a2a`, `protocols.mcp`).
 * **Static Assets:** File downloads are handled via the `artifacts` array.

### Summary

This PR introduces the foundational draft for the **AI Card** and **AI Catalog** specifications, creating a unified framework for describing and discovering AI services.

#### Key Changes

1. **Unified AI Card Schema (`ai-card-cddl.md`):**
   - **Core Abstraction:** Establishes a generic ai card format for "AI Service" entity, consolidating metadata like `id`, `name`, `description`, and `publisher` at the top level.
   - **Extensible Interfaces:** Defines an `interfaces` map, keyed by protocol type (e.g., `"a2a"`, `"mcp"`). This allows a single card to describe multiple interaction protocols, with protocol-specific details encapsulated in a `protocolSpecific` block.
   - **Trust & Identity:**
     - The `trust` section is optional, including fields for `privacyPolicyUrl` and `termsOfServiceUrl`.
     - Introduces `Attestation` field using a reference pattern (`uri`, `mediaType`, `digest`) for both publisher identity proofs and compliance evidence, supporting remote and inline credentials.
   - **Additional Metadata:**
     - Includes `maturity` to indicate lifecycle stages (`"preview"`, `"stable"`, `"deprecated"`).
     - Includes an optional `signature` field for card integrity (detached JWS).
     - Includes `identityType` hints for `id` fields (e.g., `"did"`, `"spiffe"`).

2. **AI Catalog for Discovery (`ai-catalog-cddl.md`):**
   - Introduces a lightweight catalog format, designed to be served from a well-known URL (`/.well-known/ai-catalog.json`).
   - Defines `HostInfo` for catalog metadata and `AIServiceEntry` objects for lightweight service discovery, linking to full AI Cards via `cardUrl`.

3. **Examples and Documentation:**
   - Includes `ai-card-example.md` and `ai-catalog-example.md` provide complete JSON examples, showcasing dual-protocol agents and catalog files.
   - A `comparison.md` table maps fields from previous A2A and MCP card proposals to the unified structure, emphasizing the "lift common fields, delegate protocol specifics" design.

#### Summary
This PR proposes a comprehensive, extensible, and trust-oriented foundation for an interoperable AI service ecosystem. It unifies metadata standards, simplifies multi-protocol support, and introduces a discovery mechanism to streamline integration workflows.

This table below shows how the fields from A2A and MCP map to the unified AI Card structure, based on A2A/MCP:
  - A2A Agent Card spec: https://a2a-protocol.org/latest/specification/#55-agentcard-object-structure
  - MCP server cards: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649

| Concept | A2A `AgentCard` (Spec) | MCP `Server Card` (Proposal) | **Unified `AI Card` (V1)** |
| :--- | :--- | :--- | :--- |
| **Identity (Subject)** | Implied by Host | `serverInfo.name` | **`id`** (Root URI) + `identityType` |
| **Name** | `name` | `serverInfo.title` | **`name`** |
| **Description** | `description` | `description` | **`description`** |
| **Logo** | `icon_url` | `iconUrl` | **`logoUrl`** (Rec: Data URI) |
| **Publisher** | `provider` (Object) | `serverInfo` | **`publisher`** (Object with `id`, `name`) |
| **Trust / Security** | `signatures` (Optional) | Not Defined | **`trust`** (Optional Object: Proofs, Policies) |
| **Protocol Support** | Implied (A2A Only) | Implied (MCP Only) | **`protocols`** (Map: `"a2a": {...}`, `"mcp": {...}`) |
| **Endpoint URL** | `url` | `transport.endpoint` | `protocols[x].protocolSpecific` (**Delegated**) |
| **Authentication** | `securitySchemes` | `authentication` | `protocols[x].protocolSpecific` (**Delegated**) |
| **Capabilities** | `skills` | `tools`, `resources` | `protocols[x].protocolSpecific` (**Delegated**) |
| **Versioning** | `version` | `serverInfo.version` | **`specVersion`** (Card Ver) + `updatedAt` |

#### AI-catalog example:
```json
{
  "$schema": "https://agent-card.github.io/spec/sep-0015/catalog-schema.json",
  "specVersion": "1.0",
  "host": {
    "name": "Acme Services Inc.",
    "id": "did:example:org-acme-corp",
    "documentationUrl": "https://docs.acme-corp.com/agents"
  },
  "records": [
    {
      "id": "did:example:agent-finance-001",
      "name": "Acme Finance Agent",
      "description": "Multi-protocol finance agent.",
      "tags": [
        "finance",
        "trading"
      ],
      "cardUrl": "https://api.acme-corp.com/cards/acme-finance-agent.json",
      "updatedAt": "2026-02-22T16:30:00Z"
    },
    {
      "id": "urn:example:data:market-dataset-2026q1",
      "name": "Market Dataset Q1 2026",
      "description": "Artifact-focused card for training and evaluation.",
      "tags": [
        "dataset",
        "finance"
      ],
      "cardUrl": "https://api.acme-corp.com/cards/market-dataset-2026q1.json",
      "updatedAt": "2026-02-22T16:00:00Z"
    }
  ]
}
```

### AI-card example
```json
{
  "$schema": "https://agent-card.github.io/spec/sep-0015/schema.json",
  "specVersion": "1.0",
  "cardVersion": "2026.02.22",
  "identifier": "urn:example:agent-finance-001",
  "displayName": "Acme Finance Agent",
  "description": "Executes finance workflows through multiple protocol adapters.",
  "publisher": {
    "id": "did:example:org-acme",
    "name": "Acme Financial Corp",
    "identityType": "did"
  },
  "protocols": {
    "a2a": {
      "id": "protocol/a2a",
      "version": "0.1.0",
      "required": true,
      "data": {
        "protocolVersion": "0.3.0",
        "locators": [
          {
            "type": "https",
            "role": "api",
            "uri": "https://api.acme-finance.com/a2a/v1"
          }
        ],
        "skills": [
          {
            "id": "skill-stock-analysis",
            "name": "Run Stock Analysis"
          }
        ]
      }
    },
    "mcp": {
      "id": "protocol/mcp",
      "version": "0.1.0",
      "required": true,
      "data": {
        "protocolVersion": "2025-06-18",
        "locators": [
          {
            "type": "https",
            "role": "api",
            "uri": "https://api.acme-finance.com/mcp/v1"
          }
        ],
        "capabilities": {
          "tools": {
            "listChanged": true
          }
        }
      }
    }
  },
  "trust": {
    "id": "did:example:agent-finance-001",
    "identityType": "did",
    "trustSchema": {
      "id": "https://trust.acme-finance.com/schemas/ai-card-trust-profile",
      "version": "1.0",
      "governanceUri": "https://trust.acme-finance.com/policies/ai-card-trust",
      "verificationMethods": [
        "did",
        "x509",
        "dns-01"
      ]
    },
    "privacyPolicyUrl": "https://acme-finance.com/legal/privacy",
    "termsOfServiceUrl": "https://acme-finance.com/legal/terms",
    "provenance": [
      {
        "relation": "materializedFrom",
        "sourceId": "urn:example:data:market-dataset-2026q1",
        "sourceDigest": "sha256:9999888877776666555544443333222211110000aaaabbbbccccddddeeeeffff",
        "registryUri": "oci://registry.example.com/ai-data/market-dataset@sha256:9999888877776666555544443333222211110000aaaabbbbccccddddeeeeffff",
        "statementUri": "https://trust.acme-finance.com/provenance/finance-agent-runtime-20260222.dsse.json",
        "signatureRef": "did:example:org-acme#keys-1"
      }
    ],
    "attestations": [
      {
        "type": "SOC2-Type2",
        "uri": "https://trust.acme-finance.com/reports/soc2-latest.pdf",
        "mediaType": "application/pdf",
        "digest": "sha256:a1b2c3d4"
      }
    ]
  },
  "signatures": [
    {
      "format": "jws-compact",
      "keyId": "did:example:org-acme#keys-1",
      "createdAt": "2026-02-22T16:30:00Z",
      "value": "eyJhbGciOiJFUzI1NiJ9..detached-signature"
    }
  ],
  "createdAt": "2026-02-22T16:00:00Z",
  "updatedAt": "2026-02-22T16:30:00Z"
}
```

### DIFF
--- a/specification/cddl/ai-catalog.md
+++ b/specification/cddl/ai-catalog.md
@@ -9,7 +9,7 @@ AICatalog = {
   $schema: text,              ; URI to the JSON schema
   specVersion: text,          ; Version of the Catalog spec (e.g. "1.0.0")
   ? host: HostInfo,           ; Who runs this server/registry
-  records: [* CatalogEntry]     ; The list of available AI services
+  records: [* CatalogEntry]     ; The list of available AI Catalog entries
 }
 
 HostInfo = {


mindpower:
@ToddSegal @dsp-ant Please review and let me know if you have any concerns or questions :-)

Fannon:
@mindpower thanks for writing this up! 

I have a question: Will there then still be a standalone MCP Server Card (related https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649) or would we then always use the AI Card to describe an MCP Server in full? 

With A2A it seems to work differently, because A2A already has a standalone A2A Card format, which may best be linked and not repeated?

type/ai-card-example.md
```diff
@@ -0,0 +1,111 @@
...
+    {
+      "type": "mcp",
+      "name": "Finance Agent (MCP Interface)",
+      "endpoint": "https://api.acme-finance.com/mcp/v1",
+      "authentication": {
+        "type": "http",
+        "scheme": "bearer",
+        "bearerFormat": "JWT"
+      },
+      "protocolSpecific": {
```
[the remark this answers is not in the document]
mindpower: The auth part will be identical to the MCP card. But we can further discuss the auth for A2A and MCP and other protocol. If it's hard to merge, we can keep them separately as protocol specific.

> If there is still an MCP Server card, could it be linked instead?

The MCP (and also A2A) protocol specific definition are not part of the core ai card and should be defined by each spec (A2A, or MCP). But ai-card.json should include all the data, like the following, with service declare the service (A2A, MCP) specific properties:

    + <common ai-card fields>
    + <service interfaces>
         + <service 1>
         + <service 2>
         + <service 3>

type/ai-card-example.md
```diff
@@ -0,0 +1,111 @@
...
+        "protocolVersion": "2025-06-18",
```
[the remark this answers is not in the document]
mindpower: Which specific MCP server field do you think need to be an array here?

The services here is an array which will supports multiple service interfaces. Also the ai catalog will support multiple agent declaration as well.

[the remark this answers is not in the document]
Fannon: ProtocolVersions. It's `supportedProtocolVersions` in the MCP Registry spec I think.

Reason is that one server can (and has to) support multiple protocol versions at the same time to keep compatibility with clients.

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * A short, human-readable description of the agent's purpose.
+   */
+  description: string;
+
+  /**
+   * A direct URL to the agent's logo image.
+   */
+  logoUrl?: string;
```
darrelmiller: Enterprises are likely not going to want to dereference URLs to retrieve this content.  It would be good to clarify that data URLs are supported here. https://www.rfc-editor.org/rfc/rfc2397

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * The human-readable name of the publisher.
+   */
+  name: string;
+
+  /**
+   * A URL to a verifiable credential (e.g., JWT) proving the publisher's identity.
+   */
+  attestation?: string;
```
darrelmiller: What is the reason for using just a string here, but using the Attestation type under Trust?  Could the Attestation type be reused here?

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+   * A direct URL to the agent's logo image.
+   */
+  logoUrl?: string;
+
+  /**
+   * A list of keywords to aid in discovery.
+   */
+  tags?: string[];
+
+  /**
```
darrelmiller: We should consider how to handle "terms of Service" and "privacy" URLs.

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+   * The version of the AI Card specification itself.
+   */
+  specVersion: string;
+  
+  /**
+   * The primary verifiable ID for the agent (e.g., DID).
+   */
+  id: string;
```
darrelmiller: Is it a requirement that this is "verifiable" or would that just be a recommendation?

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * An ISO 8601 timestamp of when the agent was first published.
+   */
+  createdAt: string; 
+
+  /**
+   * An ISO 8601 timestamp of when this card was last updated.
+   */
+  updatedAt: string;
```
darrelmiller: Is that just the contents of the common AI Card part, or also any part of the baseServices?  I'm assuming the latter, but just want to be sure.

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+   */
+  endpoint: string;
+
+  /**
+   * The authentication mechanism for this endpoint.
+   * (Using 'any' for simplicity, can be similar to A2A SecurityScheme).
+   * Add here if we want to standardize authentication across all services. 
+   * Otherwise, each protocol would define its own. 
+   */
+  authentication: any;
```
darrelmiller: I think it is going to be challenging to get alignment on authentication mechanisms across all protocols.  Unless we assume that everything is going to use HTTP based auth. Even then it is going to be hard.

type/src/ai-card.ts
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * A human-readable name for this specific service endpoint.
+   * For example, "Travel Agent A2A Endpoint".
+   */
+  name: string;
+
+  /**
+   * The full URL to the service endpoint.
+   */
+  endpoint: string;
```
darrelmiller: A2A has multiple URLs for a single "AI service".  I don't think people are going to want to duplicate the skills definitions for each protocol binding.

type/ai-catalog-example.md
```diff
@@ -0,0 +1,36 @@
...
+
+```json
+{
+  "$schema": "https://a2a-protocol.org/ai-catalog/v1/schema.json",
+  "specVersion": "1.0.0",
+  "host": {
+    "name": "Acme Services Inc.",
+    "id": "did:example:org-acme-corp"
+  },
+  "agents": [
```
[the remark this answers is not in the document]
mindpower: That is a fantastic optimization. Adding `updatedAt` to the catalog entry allows crawlers and registries to implement 'incremental syncing' logic.

Instead of blindly fetching every `ai-card.json `to check for changes, a client can simply compare the timestamp in the catalog with its last crawl. This significantly reduces bandwidth and processing load for both the host and the client. I will add it to the `AgentEntry` definition.

type/src/ai-card.ts
mindpower: That is a valid point. While we could use `metadata` conventions and some predefined keys, Privacy Policies and Terms of Service are fundamental trust signals required by almost every enterprise registry and app store.

Instead of burying them in a generic map, I propose we add **`privacyPolicyUrl`** and **`termsOfServiceUrl`** as explicit optional fields inside the **`Trust`** object. This groups all legal, security, and identity guarantees in one place, making it much easier for clients to programmatically verify an agent's compliance posture."

mindpower:
> I may not have fully grasped the exact intent of this project yet, but based on the README, wouldn't the `AICard` schema also serve the role that the `AICatalog` in this PR is trying to fulfill? In other words, I thought the scope of `AICard` includes server-specific metadata, such as the list of endpoints that expose `AgentCard`s, effectively providing the identities of AI `services`, rather than just defining a common schema for `AgentCard` or `MCPServerCard` (which corresponds to the `AICatalog` in this PR).


That is a fundamental architectural question, and I’m glad you raised it.
The distinction comes down to **Cardinality** and **Identity Scope**:
   1.  **`AICard` (One Identity, specific Agent):** This is designed to be the metadata for a **single, logical Agent** (e.g., 'Travel Helper'). It binds a single Verifiable Identity (DID) to a set of capabilities and services protocol. This allows granular trust: you can verify *this specific agent* is HIPAA compliant without needing to trust the entire server or other agents hosted alongside it.
  2.  **`AICatalog` (Many Identities, Discovery Layer):** This acts as the **Service Discovery** layer for a host. It solves the 'List Problem': *'I know https://www.finance-agents.com/ exists, but what agents are available there?'*

**Why separate them?**
   * **Scalability:** A Platform might host hundreds or thousands of agents. A single file containing all their metadata would be megabytes in size. The `AICatalog` allows a lightweight index (List View) that links to individual `AICard` files (Detail View).
   * **Independence:** It allows agents to be updated, versioned, and verified independently (and each agent can hash/sign their own agent card). Updating one agent's `skills` doesn't invalidate the cache for the entire server's catalog.
   * **Identity Isolation**. An AI Card (Agent) is like a Digital Passport. It binds One Identity (DID) to one set of capabilities and trusts credentials (like HIPAA/SOC2). If we put 50 agents in one file, you have a "Group Passport."
       * Signing Issues: Who signs the file? If "Agent A" is HIPAA compliant but "Agent B" is not, how do you sign the file to prove Agent A's compliance without falsely implying it for Agent B?
       * Revocation: If "Agent A" gets hacked and you need to revoke its card, you have to invalidate the entire file, which breaks trust for the other 49 legitimate agents in that file.
       * By keeping them separate, each ai-card.json is a self-contained, verifiable artifact. We can rotate Agent A's keys without touching Agent B.

ognis1205:
Sorry for the delayed response and thank you for the detailed rationale @mindpower !

I agree with the separation of `AICatalog` for a single server (a base URL with scheme + host + port) managing an index of endpoints for the identities of AI Agent-related services (A2A or MCP servers). Your explanation makes sense and I support this design.

What I was trying to clarify, however, is the specific scope of the specification managed in this repository. In other words, I wanted to understand clearly what the `AICard` is intended to define and achieve.

The current README leaves some ambiguity in terminology and scope. For example:

 - When the README mentions "server," is it referring to an A2A or MCP server, or to a single web resource identified by a base URL (scheme + host + port)?
 - Under [Core AI Card Features](https://github.com/Agent-Card/ai-card?tab=readme-ov-file#core-ai-card-features), it says the schema "may include":
   - Common server metadata (publishing company, documentation links, description, icon, server version)
   - Verifiable metadata (accreditations, certifications, reputational scores, badges)
   - Identity metadata (DID or SPIFFE)
   - Custom data defined by producers or consumers

The term "common server metadata" in the README actually made me feel that what is proposed as the `AICatalog` in this PR is exactly what the `AICard` was intended to be. Furthermore, the repository being named `ai-card` while containing `AICatalog` functionality may also contribute to this confusion.

My question is really about clarifying the intended goal and scope of AI Card itself.

mindpower:
Regarding: https://github.com/Agent-Card/ai-card/pull/4#issuecomment-3600369760

Thanks @ognis1205 ! That is an excellent nuance.

You are absolutely right that 'Agent' is too narrow, as it implies autonomy that a simple MCP tool server might not have. Using '**AI Service**' or '**AI Protocol Server**' as the umbrella term is much more precise. It correctly encompasses:
- **Autonomous Agents** (A2A)
- **Tool/Context Servers** (MCP)
- **Inference Endpoints** (OpenAPI or any other protocol)

I will adopt the suggested '**AI Protocol Server'** in the definitions to ensure we don't exclude tool-centric implementations.

I'll proceed with updating the README to reflect: (Please see: https://github.com/Agent-Card/ai-card/pull/8)
- The Terminology Shift: 'AI Service/Entity' instead of 'Agent' or generic 'Server'.
- The Scope Split: `AICard` (Identity/Capability) vs `AICatalog` (Host Inventory).

Thanks again for helping clarify these definitions. It will make the spec much more robust for implementers.

mindpower:
Thanks @OR13!  Regarding: https://github.com/Agent-Card/ai-card/pull/4#issuecomment-3604328400

This is excellent advice. I strongly agree that we want to avoid forcing every 'card verifier' to implement a kitchen sink of crypto libraries just to parse the document.

**1. Card Integrity (`signature`)** Agreed. For the root `signature` field (integrity of the card itself), we will stick to the **A2A `AgentCardSignature`** standard (JWS), as it provides a single, mandatory-to-support baseline for tamper-proofing.

**2. External Trust (`trust.attestations`)** I really like your suggestion to use a **URI + Media Type + Hash** pattern here. It decouples the *format* of the proof (JWT, PDF, C2PA, etc.) from the *structure* of the card.

Instead of defining specific fields like `credentialValue` vs `credentialUrl`, we can simplify the `Attestation` object to be a standard **Reference**:

```ini
Attestation = {
  type: text,                 ; The semantic type (e.g. "SOC2-Type2", "HIPAA-Audit")
  url: text,                  ; Location of the credential
                              ; - Remote: "https://trust.example.com/report.pdf"
                              ; - Inline: "data:application/jwt;base64,eyJ..."
  mediaType: text,            ; The format (e.g. "application/jwt", "application/pdf")
                              ; Tells the verifier which parser to use.
  ? digest: text              ; Optional integrity check (e.g. "sha256:...")
                              ; Critical for remote URLs to ensure content hasn't changed.
  ? size: uint,               ; Size in bytes.
                              ; Helps clients decide whether to download.
  ? description: text         ; Human-readable label (e.g. "2025 Security Audit")}
}

```

This allows:
  * **Simple Verifiers:** Can just check the hash and ignore formats they don't understand.
  * **Complex Verifiers:** Can use the `mediaType` to dispatch to the correct parser (e.g., verify a JWT or check a PDF signature).

I will update the spec to reflect this cleaner '**Reference**' model.

type/ai-card-cddl.md
```diff
@@ -0,0 +1,77 @@
...
+  ? tags: [* text],           ; List of keywords to aid in discovery
+  ? maturity: MaturityLevel,  ; Lifecycle stage of the agent
+  ; --- Ownership & Trust ---
+  publisher: Publisher,       ; Information about the entity that owns this agent
+  ? trust: Trust,               ; Security and compliance proofs
+  ? signature: text,          ; Detached JWS signing the card content
+  
+  ; --- Protocols ---
+  ; Map of supported protocols (e.g. "a2a" => Service, "mcp" => Service)
+  services: { * ServiceType => BaseService },
```
ognis1205: I would like to add one point that has been on my mind regarding the `services` field and the conceptual model of the `AICard`.

The A2A specification and the current MCP card proposal/draft both describe their cards as a "business card for a server", meaning a protocol-specific identity document.

But if an `AICard` contains multiple protocol entries under `services` (e.g., "a2a", "mcp", etc.), then the card no longer represents the business card of anything in particular. It effectively becomes a meta-card, a directory entry that aggregates protocol-specific service descriptors.

This ambiguity is actually one of the reasons why I felt that what is proposed as the `AICatalog` in this PR is very close to what the `AICard` itself was implicitly trying to become. If a single card can represent multiple protocols, then in practice it behaves like a catalog record rather than a protocol-scoped identity document.

To me, this raises an open design question:
 - Should an `AICard` remain a protocol-specific business card (1 protocol = 1 card)?
 - Or should the `AICard` explicitly become a meta-card, and the `AICatalog` contain protocol-specific cards?

I think clarifying this distinction explicitly would help solidify the semantics of both `AICard` and `AICatalog`.

mindpower: These are excellent architectural questions. Here is how I see the scope of `AICard`. 
Regarding the "meta-card" concern: I view the `AICard` as an **Identity Document** for a logical entity, rather than just a config file for a protocol.

**The Analogy**: Think of a human Business Card. It lists a **Phone Number**, an **Email Address**, and a **Slack Handle**.

- It supports "multiple protocols" (Voice, SMTP, Chat).

It does _not_ become a "directory" or "phone book" because of this. It is still a single card describing **one person**.

**The Benefit**: By grouping A2A and MCP interfaces under one `AICard`, we allow **Single Verification**. A client verifies the Agent's DID once (checking the signature/compliance), and that trust automatically extends to all its interfaces (A2A and MCP). If we enforced "1 Protocol = 1 Card," we would fragment the identity, forcing clients to perform redundant trust checks for every single endpoint, which is brittle and computationally expensive.

type/ai-card-cddl.md
```diff
@@ -0,0 +1,77 @@
+### AI Card Specification in CDDL format
+```ini
+; ============================================================================
+; AI Card Specification (v1)
+; A unified metadata format for AI Agents
```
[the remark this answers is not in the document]
mindpower: Please see the other reply below.

mindpower:
Just updated `ai-card` and `ai-catalog` specs and examples in this PR based on all the discussions so far:
- [ai-card-cddl](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-card-cddl.md)
- [ai-catalog-cddl](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-catalog-cddl.md)
- [ai-card-example](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-card-example.md)
- [ai-catalog-example](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-catalog-example.md)

Also created [PR #8](https://github.com/Agent-Card/ai-card/pull/8) for the README change to clarify all the terminology, scopes, and architecture models.

type/ai-catalog-cddl.md
```diff
@@ -0,0 +1,30 @@
...
+; ============================================================================
+; AI Catalog Specification (v1)
+; The "Index" file served at /.well-known/ai-catalog.json
+; ============================================================================
+
+AICatalog = {
+  $schema: text,              ; URI to the JSON schema
+  specVersion: text,          ; Version of the Catalog spec (e.g. "1.0.0")
+  host: HostInfo,             ; Who runs this server/registry
+  agents: [* AICardEntry] ; The list of available agents
```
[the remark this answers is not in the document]
ognis1205: The points you raised are absolutely correct.

I agree that `AI Service` and `interfaces` are much better aligned with a protocol-agnostic design. Thank you for the thoughtful suggestion. I fully support adopting this terminology!

type/ai-card-cddl.md
ognis1205: Thanks @mindpower ! I completely agree with both points:

> Regarding the "meta-card" concern: I view the `AICard` as an Identity Document for a logical entity, rather than just a config file for a protocol.
> **The Benefit**: By grouping A2A and MCP interfaces under one AICard, we allow **Single Verification**.

I think the overall direction makes sense.

That said, I still feel we need to explore the question of what exactly this "logical entity" corresponds to in real deployments. For example, is an `AIService` always expected to expose exactly one MCP server or one A2A server? Or could a single `AIService` operate multiple MCP endpoints (e.g., one for tools and one for workflows) or multiple A2A servers (e.g., different domains or capabilities)?

If such patterns are valid, and I suspect they might be in production settings, then the mapping between the "single logical entity" and the underlying protocol servers becomes less clear. In that case, it would be important to clarify whether the `AICard` is representing:

1. a single agent instance,
2. a product-level service composed of multiple protocol servers, or
3. a higher-level organizational identity grouping many protocol endpoints.

Here is a concrete example:

> A company exposes an "AI Document Assistant" as one logical service. Internally, it runs
> - two MCP servers (one for document-editing tools, one for knowledge-base queries),
> - one A2A server for agent-to-agent orchestration,
> all representing the same "AIService" to clients.

This doesn't seem far-fetched, but it would complicate assumptions about "one logical entity -> one server per protocol." So I agree with the single-verification direction, but I also believe we need a bit more clarity on the expected granularity of an `AIService` and how it maps to real-world multi-endpoint setups.

type/ai-catalog-cddl.md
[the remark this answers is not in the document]
mindpower: Here is a summary of the name update and reasons:

| Field | Old Name  | New Name  | Reason for the Fix |
| -- | -- | -- | -- |
Top-Level Entity | `Agent`  | `AI Service` | Industry-standard term, covers both Agent and Tool. |
Catalog List Field | `agents` | `services` | The catalog lists available AI Services. This is the natural noun for the list of entities. |
Card Map Field | <span style="white-space: nowrap;">`services`&nbsp;or&nbsp;`interfaces` </span>| <span style="white-space: nowrap;">`protocols` </span> | If the entity is the "Service," the map cannot be "services." `protocols` is the most accurate description for the keys (A2A, MCP), which define the communication layer. |

type/ai-card-cddl.md
mindpower: Please see comments below (https://github.com/Agent-Card/ai-card/pull/4#discussion_r2612699349)

type/ai-catalog-cddl.md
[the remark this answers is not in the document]
mindpower: Thanks for the suggestion!
1. **`ProtocolDetail`:** I like this rename for the object type. It clearly signals '_This object holds the details for one specific protocol binding_.' I will adopt that.
2. **Naming the `type` field:** I prefer to keep the field inside the object as **`type`** (instead of `protocolName`).
 * **Reason:** `type` is the standard convention for discriminators in JSON schemas. Using `protocolName` might get confused with the human-readable `name` field right next to it. 
 * Illustration: Imagine the confusion if we had `protocolName` (the enum) next to `name` (the label) in the `ProtocolDetail` structure below 

```typescript
// Map
protocols: Record<ProtocolType, ProtocolDetail[]>

// Object
interface ProtocolDetail {
  type: string; // "a2a"
  name?: string; // "Primary Interface"
  protocolSpecific: ...
}
```

 3. **Human-Readable Labels:** You raised a good point about how these look to users. Even though we call the map `protocols`, the human `name` field doesn't need to be robotic.
 * We have a semantic conflict between the Technical Architecture (`Protocol`) and the User Experience (`Service/Interface`). we (correctly) renamed the schema field to `protocols` to be technically precise, but that makes the English label feel awkward.
 * The Fix: The Human-Readable Name (`name`) does not need to use the word "`Protocol`" or "`Service`." It should describe the Function.
 * *Instead of :* `"name": "Finance Agent A2A Protocol"` (sounds very **robotic**)
 * *Use:* `"name": "Finance Agent Interface"` or  `"name": "Finance Agent MCP workflow service"`
 * We can leave the `name` field flexible for developers to describe the *function* of that interface (e.g., 'Primary', 'Backup', 'Tools') rather than just repeating the protocol name.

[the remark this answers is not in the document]
ognis1205: Thanks for the clarification, I'm aligned with this direction. Looks good to proceed with this change!

ognis1205:
> Just updated `ai-card` and `ai-catalog` specs and examples in this PR based on all the discussions so far:
> 
> * [ai-card-cddl](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-card-cddl.md)
> * [ai-catalog-cddl](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-catalog-cddl.md)
> * [ai-card-example](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-card-example.md)
> * [ai-catalog-example](https://github.com/mindpower/ai-card/blob/ed1c4aac12eecf2f56da58a93334549e77187670/type/ai-catalog-example.md)
> 
> Also created [PR #8](https://github.com/Agent-Card/ai-card/pull/8) for the README change to clarify all the terminology, scopes, and architecture models. And updated the description above for this PR.

Thanks for pulling all of this together, appreciate the updates to both the specs and examples.
I'll review the README changes in PR #8 as well.

type/src/ai-catalog.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * A list of tags for filtering and discovery.
+   */
+  tags?: string[];
+  
+  /**
+   * The full, absolute URL to this AI service's complete `ai-card.json` file.
+   */
+  cardUrl: string;
```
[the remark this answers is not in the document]
mindpower: @tadasant I agree with your approach for discovery. We do not want to force an extra hop, nor do we want to force `MCP` into an envelope that contains governance/trust fields you aren't ready for today. The `ai-catalog` should simply be a discovery list that can point directly to native `MCP` Server Cards. Please see my comments below for more details about that.

ToddSegal (approved):
Approving baseline for merge. We can iterate afterwards on ensuring that card can be any type of card (mcp server card, a2a card, etc) and that we can embed the ai card metadata in the other cards.

PR #14, opened 2026-01-24
SEP 0014 - AI Card Discovery

Fannon:
This SEP proposes a standardized discovery mechanism for AI-native protocols (MCP and A2A) using a `.well-known` URI endpoint. 

It makes a strong assumption that MCP and A2A do not create a shared model, each will own its own "card" metadata files. But we can still share the discovery mechanism between the two (and potentially other / upcoming) protocols.

## Problem

Both MCP and A2A protocols have defined protocol-specific discovery endpoints that:
- Assume **one server/agent per host** (MCP: `/.well-known/mcp/server-card.json`, A2A: `/.well-known/agent.json`)
- Require clients to probe multiple endpoints for multi-protocol discovery
- Create ecosystem fragmentation

## Solution

A single, protocol-agnostic discovery endpoint at `/.well-known/ai-cards.json` that:
- Supports **multiple MCP servers and/or A2A agents** per host
-  Enables single-request multi-protocol discovery
- Maintains protocol autonomy (each protocol owns its card format)
- Uses simple JSON Schema with 5 required fields per protocol
- Supports both relative and absolute URLs

## Example

```json
{
  "$schema": "https://ai-card.org/schemas/ai-card-v0.schema.json",
  "protocols": [
    {
      "type": "mcp",
      "endpoints": [{"url": "./mcp-petstore"}],
      "metadata": {
        "type": "mcp-server-card",
        "url": "./petstore.mcp.json"
      }
    },
    {
      "type": "a2a",
      "endpoints": [{"url": "/agents/support"}],
      "metadata": {
        "type": "agent-card",
        "url": "/metadata/SupportAgent.json"
      }
    }
  ]
}

### DIFF
--- a/seps/0000-ai-card-discovery.md
+++ b/seps/0000-ai-card-discovery.md
@@ -149,6 +149,7 @@ The content and structure of metadata files referenced by `metadata.url` are **o
 - **A2A Agent Cards**: Already defined by the A2A specification. The A2A community maintains full control over this schema.
 
 This SEP **intentionally does not mandate** any particular structure for these metadata files. It only defines the discovery mechanism that points to them.
+The scope could be increased to also indicating the version(s) of the protocols and metadata files that are provided.
 
 ## Rationale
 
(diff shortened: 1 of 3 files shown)


scutuatua-crypto (approved):

muscariello:
I think this is obsolete and we should be able to close it.

PR #18, opened 2026-02-23
spec: post-meeting converged AI Card profile proposal

muscariello:
## Summary

This PR adds a new post-meeting specification proposal that converges the current AI Card draft direction into a single profile focused on:

1. Preserving protocol autonomy through delegated extension modules.
2. Supporting multiple discovery and distribution technologies.
3. Strengthening identity and cryptographic provenance guidance.
4. Providing a practical migration path from previous draft variants.

## Overview

This PR rewrites the SEP-0015 converged AI Card profile as an OCI-based specification, replacing the custom card/catalog model with one that follows OCI Image and Distribution specification. The result is that common AI metadata is natively storable, distributable, and signable in any OCI registry without any custom tooling.

## What changes

**Data model**

The top-level AI Card document becomes a standard OCI Image Manifest, identified as an AI Manifest by its artifactType (`application/vnd.aaif.ai.manifest.v1+json`). Common metadata like identity and publisher details is moved into a dedicated config blob (`application/vnd.aaif.ai.card.metadata.v1+json`). Protocol-specific metadata (A2A, MCP) and data asset references become typed OCI layer blobs, with artifactTypes owned by this spec but content schemas owned by the respective upstream projects. The AI Catalog becomes a standard OCI Image Index.

The data model can be expressed using a MerkleDAG:
- AI Manifest is a node linking common AI card metadata with protocol-specific layers
- AI Catalog is a node that aggregates multiple AI Manifests for discovery
- Referrers are nodes linking external data to AI Manifests for application-layer logic

```mermaid
graph TB
  subgraph Catalog["OCI Index"]
    AICatalog["AI Catalog"]
  end

  AICatalog --> AIM1["AI Manifest 1"]
  AICatalog --> AIMn["AI Manifest N"]

  subgraph Manifest1["OCI Manifest"]
    AIM1 --> Config1["AI Card Metadata"]
    AIM1 --> Layer1A["A2A Card"]
    AIM1 --> Layer1B["AGNTCY DIR Record"]
    Config1 ~~~ Layer1A ~~~ Layer1B
  end

  subgraph ManifestN["OCI Manifest"]
    AIMn --> ConfigN["AI Card Metadata"]
    AIMn --> LayerN1["MCP Server"]
    AIMn --> LayerN2["HuggingFace Model"]
    ConfigN ~~~ LayerN1 ~~~ LayerN2
  end

  subgraph AppLayer["OCI Referrers"]
    Ext1["Signature"]
    Ext2["Identity Claims"]
  end

  Ext1 --> AIM1
  Ext2 --> AIMn

  style Catalog fill:#dbeafe,stroke:#3b82f6,color:#1e3a5f
  style Manifest1 fill:#dcfce7,stroke:#22c55e
  style ManifestN fill:#dcfce7,stroke:#22c55e
  style AppLayer fill:#fef3c7,stroke:#f59e0b
```

**Signing**

The embedded `signatures` array and the associated subject commitment model are removed entirely. Signing and attestation are handled via the OCI Referrers API using cosign or notation, following the same pattern already used for signing container images.

**Discovery**

A new statically hosted registry binding is defined, rooted at `/.well-known/ai-registry`. This implements the read-only subset of the OCI Distribution Spec (manifest fetch, blob fetch, tags list, referrers list). For static only serving, `_catalog.json` is defined which serves AI Catalog file (OCI Index). Dynamic autodiscovery is also supported through a common subject descriptor included in the AI Manifest and works out-of-box with existing OCI registries.

**Spec document**

The document is refactored from a proposal into a normative specification: RFC 2119 conformance language, an Abstract section, a Related Work section referencing CNCF ModelPack/AGNTCY DIR, removal of Open Questions, and updated conformance levels (L0–L3).


mindpower (commented):
Hey @muscariello , thanks for putting this converged view together! I'm really glad to see we are completely aligned on the core structural objects here - the Publisher, Trust, and Attestation blocks look great and match the https://github.com/Agent-Card/ai-card/pull/4 proposal perfectly.

I also completely agree with your addition of the artifacts array. Having a dedicated way to reference static data (like model weights, datasets, or static skill files) is a great addition.

My main piece of feedback is regarding the transport/endpoint layer. I notice this drops the protocols map from PR #4 in favor of moving locators, artifacts, and modules to the root level alongside a cardKind enum (live-service, data-asset, hybrid).

My concern is that flattening this creates ambiguity for routers / clients, especially with the hybrid designation. If an agent exposes both an MCP protocol and an A2A protocol, a flattened locators array forces the routing client to loop through every URL and guess which endpoint belongs to which protocol based on the URI or media type. It breaks the strong typing we achieved in the core objects.

Could we look at a 'best of both' approach?
1. Keep the artifacts: [* Artifact] array at the root level for static data downloads (the 'nouns').
2. Restore the protocols: { "mcp": {...}, "a2a": {...} } map for live endpoints (the 'verbs'). This gives each protocol a clean, isolated namespace.
3. Drop the cardKind enum and the generic locators array. If a card has artifacts, we know it has data assets. If it has protocols, we know it's a live service. If it has both, it's natively a 'hybrid' without needing a string enum to tell the router to guess.

This keeps the strict typing you aimed for, but ensures the transport layer remains cleanly isolated. What are your thoughts on combining these approaches?

mindpower:
> The reason I added the card kind was to manage more general cases where you have an MCP card as data asset and also as live services. The current MCP registry is about data assets with a mix of live services too. To learn more about the kind of card you need to scan inside, and therefore you cannot filter early.
> 
> I am not sure my intent was to flatten, I am mostly concerned about discriminating live vs data. The main reason for that is that the security model is very different for these two use case: one is authentication the other is mostly provenance.

Ah, that makes complete sense for the registry layer . Having `cardKind` for that O(1) early filtering so the registry doesn't have to parse the whole payload is a great optimization.

I would gently push back on one nuance, though: in a zero-trust Agentic world, live services *also* heavily rely on provenance. Authentication alone isn't enough for an agent API - we still need supply-chain provenance (like AI-BOMs, signed build manifests, or base model attestations) to actually trust the live inference endpoint.

But your core point about early filtering sounds right. What if we combine both of our goals to satisfy both the Registry and the router?

1. **Keep `cardKind` at the root:** This gives the registry the exact early-filtering index it needs to know what it's dealing with before parsing the payload.
2. **Replace the flat `locators` array with `protocols` (for live APIs) and `artifacts` (for static data):** This fixes the routing ambiguity. If the router pulls a `live-service`, it instantly grabs `protocols.a2a` or `protocols.mcp` without having to loop through generic locators to guess the transport. If it pulls a `data-asset`, it just grabs the `artifacts` array.

This way, the Registry gets its fast filtering, the Router gets its strict typing without guessing, and the unified `Trust` block at the root handles provenance for both use cases.

Does that sounds good to you?

mindpower:
This latest commit looks fantastic, @muscariello ! We are practically 100% converged. Dropping the rigid `cardKind` and bringing back the map structure completely solves the routing ambiguity we discussed. Combined with the `tags` you included earlier, this perfectly aligns with the extensible routing and registry discovery goals from the original PR #4 proposal.

**I just have one minor structural thought, and a quick procedural suggestion:**

**1. The `modules` wrapper:** Do we gain much by nesting `protocols` and `artifacts` inside a generic `modules` object? (i.e., `card.modules.protocols.a2a` vs `card.protocols.a2a`). Usually, keeping primary domain concepts at the root level reduces parsing friction for clients. If we just promote `protocols` and `artifacts` directly to the root (right next to `tags` and `publisher`), it keeps the JSON incredibly clean and flat.

**2. Preserving the Working Group History:**
Since this converged schema now beautifully reflects the isolated namespaces and core identity blocks we hammered out in PR 4 (plus the excellent addition of the `artifacts` array!), how do you want to handle the merge? PR #4 currently has 70+ comments of deep architectural history and consensus-building from the working group and the broader open-source community.

Would you be open to pushing this final, converged CDDL (with the flattened root) directly into PR #4? That way, we keep all the rich historical context and decision logs attached to the final merge, rather than fragmenting it.

Either way, incredible work pulling all these threads together! Let me know what you think.

specification/examples/converged-data-asset-card.json
```diff
@@ -0,0 +1,59 @@
...
+    "id": "did:example:org-acme",
+    "identifierType": "did",
+    "name": "Acme Financial Corp"
+  },
+  "domains": [
+    "finance"
+  ],
+  "modules": {
+    "artifacts": {
+      "dataset": {
```
darrelmiller: What is this key? And how is it different than the Id of the artifact?

mindpower:
Hey @muscariello , just a quick update: Based on discussions https://github.com/Agent-Card/ai-card/pull/18#issuecomment-4006560855.  I've pulled the converged CDDL into PR #4! I went ahead and flattened the protocols and artifacts up to the root (dropping the modules wrapper) as we discussed, and I also added you as a Co-Author on the commit and also updated you as an Assignee on that PR so we can drive it across the finish line together. Whenever you are ready, feel free to close this PR out, and we can continue the final naming/polish discussions over on #4 where all the historical context lives. Thanks again for driving this convergence.

Tehsmash:
>  For many of those developers, standing up an OCI registry or understanding Image Manifests just to share a simple agent card introduces too much friction. 

In addition to what @muscariello said, most developers should never need to interact directly with the proposed structure unless they are building a CLI, library or registry on top of the spec which needs to understand the structure in order to navigate it.

For most agentic developers there should be tooling, libraries, or CLIs to interact with the ai-card catalog/registry whether static or dynamic. For example:

If you're using modelpack to publish a model you create a Modelfile (like a Dockerfile) and go through modctl which packs the data into the OCI structure and uses ORAS underneath. You also use modctl to pull from the registry static local or remote and extract the model back into a directory.

I envision the ecosystem will be similar for AI Card as for modelpack, and if we use the same structure for static and for dynamic catalogs then the same tooling can be used for both at least from a reader client's perspective.

PR #36, opened 2026-05-13
spec: adopt `urn:air:` naming standard for AI artifacts (ADR 0015)

mindpower:
This pull request introduces **ADR 0015**, proposaling a standardized, domain-anchored URN naming convention (`urn:air:`) for all AI actors and artifacts within the AI Catalog ecosystem. Evolving from the recommendations in [PR #19](https://github.com/Agent-Card/ai-catalog/pull/19), this proposal ensures global federation, consistent indexing, and industry-wide interoperability across registries and orchestrators.

### 1. Architectural Decision Record: `adr/0015-agent-identifier-naming.md`
- **Formalized Naming Contract**: Defines the universal URN structure: `urn:air:{publisher}:{namespace}:{name}`.
  - `publisher`: The anchoring domain where the artifact is hosted (e.g., `example.com`).
  - `namespace`: Optional, hierarchical segments separated by `:` (e.g., `hr`, `finance`, `finance:agent`, `mcp`, `skill`, `catalog`).
  - `name`: The specific, distinguishing name of the artifact.
- **Separation of Concerns**: Establishes a clear boundary between **logical discovery names** (the stable `identifier` URN) and **cryptographic trust identities** (the `identity` field in `trustManifest`, which may use DIDs, SPIFFE, x509, etc.).

### 2. Core Specification: `specification/ai-catalog.md`
- **Specification Text Update**: Updated the `identifier` member definition to mandate the `urn:air:` format. (**air->ai resource**)
- **Example Alignment**: Updated 38 example JSON blocks and narrative references across the specification to reflect the new standard, including:
  - A2A Agent Cards (`urn:air:acme.com:agent:finance`)
  - MCP Server Cards (`urn:air:io.modelcontextprotocol.anonymous:mcp:brave-search`)
  - Nested & Hierarchical Catalogs (`urn:air:acme.com:catalog:finance`)
  - Dual-Protocol (MCP + A2A) Wrappers (`urn:air:acme.com:agent:finance:mcp`)
  - Security Typosquatting Scenarios (`urn:air:acme.com:agent:financ` vs `urn:ai:acme.com:agent:finance`)

## Architectural Rationale:
- **Nomenclature Stability**: The URN acts as an abstract, stable contract. Relocating workloads or changing underlying cloud infrastructure does not break client discovery or routing logic.
- **Global Uniqueness**: Anchoring URNs to fully qualified domain names (FQDNs) guarantees cross-network uniqueness without central registration overhead.
[...]

PR #37, opened 2026-05-15
Spec: Update mediaType to type for catalogEntry and add known types

mindpower:
### Move from `mediaType` to `type` in `CatalogEntry`
- **Decision**: Renamed the `mediaType` field to `type` in catalog entries to avoid confusion with generic IANA media types.
- **Open Format with Known Types**: Clarified that the `type` field is an open text format accepting any string value. However, a set of recognized "known types" is recommended for interoperability (e.g., `application/a2a-agent-card+json`, `application/mcp-server+json`, `application/ai-catalog+json`, `application/agent-registry`, and format-specific skills like `+gzip`, `+zip`, `+json`). Clients are responsible for handling new or unknown types.
- **ADR Created/Updated**: Added and updated [ADR 0014](file:///Users/jbu/Development/ai-card/adr/0014-media-type-to-type.md) to document this decision.


adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Decision
+We will rename the `mediaType` field to `type` in the `CatalogEntry` schema.
+The `type` field is an open text format, so any string value is accepted. The following are recognized "known types" in the ecosystem:
+- `application/a2a-agent-card+json`
+- `application/mcp-server+json`
+- `application/ai-catalog+json`
+- `application/ai-skills+tgz`
+- `application/ai-skills+zip`
+- `application/ai-skills+md`
```
darrelmiller: As there is no registered suffix, it would not be possible at present to register this media type.

adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,31 @@
...
+
+Furthermore, we need to support combined types that specify both the artifact type and the payload format (e.g., `application/ai-skills+tgz`).
+
+## Decision
+We will rename the `mediaType` field to `type` in the `CatalogEntry` schema.
+The `type` field is an open text format, so any string value is accepted. The following are recognized "known types" in the ecosystem:
+- `application/a2a-agent-card+json`
+- `application/mcp-server+json`
+- `application/ai-catalog+json`
+- `application/ai-skills+tgz`
```
darrelmiller: There is a +gzip but there is not +tgz that is registered and so currently it would not be possible to register this media type.

specification/ai-catalog.md
```diff
@@ -197,29 +197,30 @@ It MUST contain the following members:
...
-  - `application/mcp-server-card+json` — an MCP Server Card
-  - Any other media type defined by a protocol specification (e.g.,
-    `application/agentskill+zip` for skill definitions)
+  - `application/mcp-server+json` — an MCP Server Card
+  - `application/ai-catalog+json` — a nested AI Catalog
+  - `application/ai-skills+tgz` — an AI Skill bundle (compressed tarball)
+  - `application/ai-skills+zip` — an AI Skill bundle (ZIP archive)
+  - `application/ai-skills+md` — an AI Skill defined in Markdown
+
+  These values combine the artifact type with the payload format. For any new or custom types not listed here, it is up to the specific client implementation to handle them correctly.
```
Tehsmash: Should we split the ai-catalog/ai-card managed media types, from the third party ones? I was envisioning something like:

```
AI Catalog types:
application/ai-catalog+json - nested AI Catalog
application/agent-card+json - reserved Agent Card media type

Third Party Types:
(A2A, MCP etc)
```

mindpower: Sounds good. Updated.

adr/0014-media-type-to-type.md
mindpower: Updated the adr and spec. Added the +gzip but removed +tgz.

adr/0014-media-type-to-type.md
mindpower: Changed this `application/ai-skills+md` to `text/markdown; profile=ai-skill`

jdamick:
The attestation object already has a "type" and "mediaType" and now type in the entry is actually a media type. It seems confusing and inconsistent to me when i read this again.

jdamick:
lgtm

specification/ai-catalog.md
```diff
@@ -197,29 +197,34 @@ It MUST contain the following members:
...
+
+  **Integrated Ecosystem & Third-Party Types (Governed externally):**
   - `application/a2a-agent-card+json` — an A2A Agent Card
-  - `application/mcp-server-card+json` — an MCP Server Card
-  - Any other media type defined by a protocol specification (e.g.,
-    `application/agentskill+zip` for skill definitions)
+  - `application/mcp-server+json` — an MCP Server Card
+  - `application/ai-skills+zip` — an AI Skill bundle (ZIP archive)
+  - `application/ai-skills+gzip` — an AI Skill bundle (gzipped tarball)
+  - `text/markdown; profile=ai-skill` — an AI Skill defined in a standard Markdown file
```
darrelmiller: Profiles are expected to be a URI https://datatracker.ietf.org/doc/html/rfc6906#section-3

mindpower: Thanks for the information. Yes you're right. In that case the type looks a little long here. Let me summarize different options below.

mindpower:
We have been discussing for a while. Here are the requirements and context: basically we need to tell both "what this is logically" (an AI Skill) with "how it is encoded" (Json, ZIP, GZIP, Markdown) into a single `type` field, that's where we run into the various issues we discussed in the meeting and also comments in PR #36 and this PR #37.

Combing "What is logical" + "how is it encoded" ==> 
   - a2a agents, mcp server, which is a json file
   - Another catalog, which is also a json file
   - An agent registry (rest APIs), which is an endpoint / url, it has no MIME type (encoding),  
   - an **AI Skill** is a physical payload (a file or bundle) that the client has to download and parse. It can be a markdown file, zip, or gzip. The clientneed know whether it's dealing with a ZIP file, a gzipped tarball, or a raw Markdown file before it tries to open it.

Here are a summary of the different options to solve this, ranked from the cleanest architectural pattern to the most pragmatic workaround:

### 1. The "Two-Field" Approach (Cleanest Architecture)

The standard way to resolve this tension is to separate the concerns into two distinct fields: one for the logical domain concept, and one for the transport format.

* **`type`** (or `kind`): Describes the logical entity.
* **`format`** (or `mediaType` / `contentType`): Describes how the bytes are encoded, using strict, standard IANA MIME types.

**Example:**

```json
{
  "identifier": "urn:air:example:skill:calculator",
  "type": "ai-skill",
  "mediaType": "application/zip",
  "url": "https://example.com/skills/calc.zip"
}

```

**Why this is best:** We don't have to invent or register custom media types like `application/ai-skills+zip`. You just use standard `application/zip` or `text/markdown`, and the `type` field handles the AI-specific classification.

### 2. The "Profile" Approach (Strict HTTP Compliance)

If we must combine them into a single string that behaves like a standard media type, the most compliant way—which addresses @darrelmiller  feedback about RFC 6906 - is to use standard media types with a **`profile`** parameter.

As darrel noted in the PR, profiles should be URIs, not just bare strings like `ai-skill`.

**Examples:**

* **ZIP Bundle:** `"type": "application/zip; profile=\"urn:air:ai-skill\""`
* **Markdown Skill:** `"type": "text/markdown; profile=\"urn:air:ai-skill\""`

**Why this works:** It tells the client exactly how to parse the file (`application/zip`) but also guarantees that the contents inside that ZIP conform to the specific AI Skill schema (`profile="urn..."`).
**Issue**: This does not solve the AI registry endpoint type issue (which mime media type should we use?)

### 3. The "Structured Enum" Approach (Pragmatic Compromise)

If we strictly prefer to drop the `mediaType` field and using a single, open-text `type` field for everything (as `adr-0014` suggests), you can use suffixes but drop the `application/` prefix entirely to stop pretending they are strict IANA media types.

Instead of `application/ai-skills+zip`, just use our own namespaced enums:

* **Values:** `ai-skill+zip`, `ai-skill+gzip`, `ai-skill+md`

**Example:**

```json
{
  "identifier": "urn:air:example:skill:calculator",
  "type": "ai-skill+zip",
  "url": "https://example.com/skills/calc.zip"
}

```
The list of types are:
  - a2a-agent-card+json — A2A Agent Card (json format)
  - mcp-server+json — MCP Server Card (json format)
  - ai-skills+zip — AI Skill bundle in a ZIP archive
  - ai-skills+gzip — AI Skill bundle in a gzipped tarball
  - ai-skills+markdown - AI Skill markdown format
  - ai-catalog+json - AI catalog (json format)
  - ai-registry - Rest based AI Registry (endpoint)
 
**Why this works:** It provides the client with both pieces of information (it's a skill, and it's zipped) without violating IANA registration rules, because it's no longer formatted as a MIME type.

---

Given that `adr-0014` specifically renames the field from `mediaType` to `type` to avoid confusion with generic IANA types, **Option 3** (or **Option 1** if the we want to add second field) seems the most aligned with the PR's goals while satisfying the requirement to identify the file format.

Tehsmash:
I know we introduced the whole markdown + profile thing a few weeks ago, but I didn't really process it at the time. I Definitely do not agree with that approach as we end up with a completely different type for different ai-skills, so now when I'm searching a catalog for an ai-skill I have to look for both things prefixed with application/ai-skill and now any markdown with that profile. If we go that approach we might as well replace them all with application/json profile=a2a-card etc. 

From the OCI perspective they stick to IANA format but I'm not sure they are so strict as we're laying down here. For example in model pack https://github.com/modelpack/model-spec/blob/main/docs/spec.md you can see the various media types they have defined for packing a model into an OCI bundle. As we discussed further up `+tar` isn't natively supported in the IANA postfixes, for this reason they create sub-types, then use the postfixes which are availalble, for example:

> **application/vnd.cncf.model.weight.config.v1.raw**: The layer is an unarchived, uncompressed configuration file of the model weights like tokenizer.json, config.json, etc.

> **application/vnd.cncf.model.weight.config.v1.tar**: The layer is a [tar archive](https://en.wikipedia.org/wiki/Tar_(computing)) that contains configuration files of the model weights like tokenizer.json, config.json, etc.

> **application/vnd.cncf.model.weight.config.v1.tar+gzip**: The layer is a [tar archive](https://en.wikipedia.org/wiki/Tar_(computing)) that contains configuration files of the model weights like tokenizer.json, config.json, etc. The archive is compressed with [gzip](https://tools.ietf.org/html/rfc1952).

In this way we could also take a similar approach:

application/ai-skill.md
application/ai-skill.tar
application/ai-skill.tar+gzip

This would allow for easy filtering by `application/ai-skill.*` glob, then have further parsing after that point. 

As a follow up to this you could even think about breaking the ai-skill bundle down if you wanted i.e.

application/ai-skill.skill.md
application/ai-skill.references.tar
application/ai-skill.scripts.tar

Then you can use an ai-catalog to bundle these together to create a complete skill package. This approach would allow you to pull only the markdown to inspect it first without pulling all the components, to share the same references tar blob between two different versions of the skill etc etc.

mindpower:
Thanks, @Tehsmash . I agree - the suggested subtype + postfix approach strikes a pretty good balance across all our different requirements.

The only tweak I would suggest is dropping the application/ prefix entirely if we can. Since we are moving this field to type specifically to avoid confusion with strict IANA-registered media types, keeping the prefix feels a bit misleading. Plus, if every single type just defaults to starting with `application/`, it becomes redundant noise rather than useful information.

So, if we pick up this approach, we'd end up with values like below:
- a2a-agent-card.json — A2A Agent Card (json format)
- mcp-server.json — MCP Server Card (json format)
- ai-skills.md- AI Skill markdown format
- ai-skills.md+zip — AI Skill bundle in a ZIP archive
- ai-skills.tar+gzip — AI Skill bundle in a gzipped tarball
- ai-catalog+json - AI catalog (json format)
- ai-registry - Rest based AI Registry (endpoint)

Note: Most of `type` value here will look like simply take the last part of `url` field value.

For example:
| type | url |
|-|-|
|a2a-agent-card.json | https://www.example.com/.well-known/agent-card.json|
|mcp-server.json | https://www.example.com/.well-known/mcp-server.json|
|ai-skills.md | https://www.example.com/stock/skills/stock-skills.md|
|ai-catalog.json | https://www.example.com/.well-known/ai-catalog.json|

muscariello:
Shouldn't it be agent skills and not AI skills ?

Tehsmash:
> Shouldn't it be agent skills and not AI skills ?

+1 this is a specific format, AgentSkill.md so I think it should reflect that

> The only tweak I would suggest is dropping the application/ prefix entirely if we can. Since we are moving this field to type specifically to avoid confusion with strict IANA-registered media types

I would be cautious of dropping the prefix because while we're changing mediaType to type avoid confusion I think we still want to use the IANA format/semantics and be able to parse the types using an IANA format parser, otherwise we could just pick arbitrary strings for this type field and make it specific to AI catalog.

adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,55 @@
...
+#### Core Protocol Types (Governed by the AI Catalog WG)
+- `application/ai-catalog+json` — nested AI Catalog
+- `application/agent-card+json` — reserved for a generic Agent Card format
+
+#### Integrated Ecosystem & Third-Party Types (Governed externally)
+- `application/a2a-agent-card+json` — A2A Agent Card
+- `application/mcp-server+json` — MCP Server Card
+- `application/ai-skills+zip` — AI Skill bundle in a ZIP archive
+- `application/ai-skills+gzip` — AI Skill bundle in a gzipped tarball
+- `text/markdown; profile=ai-skill` — AI Skill defined in a standard Markdown file
```
jonathanhefner: Should these be "Agent Skill" (per https://agentskills.io/) instead of "AI Skill"?


```suggestion
- `application/agent-skills+zip` — Agent Skill bundle in a ZIP archive
- `application/agent-skills+gzip` — Agent Skill bundle in a gzipped tarball
- `text/markdown; profile=agent-skill` — Agent Skill defined in a standard Markdown file
```

specification/ai-catalog.md
```diff
@@ -117,19 +117,19 @@ For example, a minimal catalog listing three AI artifacts:
     {
       "identifier": "urn:example:skill:code-review",
       "displayName": "Code Review Assistant",
-      "mediaType": "application/agentskill+zip",
+      "type": "application/ai-skills+zip",
```
jonathanhefner: I see this _was_ `agentskill`, but was changed to `ai-skills`.  Was that intended?

specification/ai-catalog.md
jonathanhefner:
```suggestion
  - `application/agent-skills+zip` — an Agent Skill bundle (ZIP archive)
  - `application/agent-skills+gzip` — an Agent Skill bundle (gzipped tarball)
  - `text/markdown; profile=agent-skill` — an Agent Skill defined in a standard Markdown file
```

adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,55 @@
...
+To ensure that our recommended types are fully registerable as official IANA media types, we adhere strictly to the IANA Structured Syntax Suffix registry rules (RFC 6838/6839):
+- **Custom Ecosystem Formats**: Types like A2A Agent Cards (`application/a2a-agent-card+json`), MCP Servers (`application/mcp-server+json`), and Nested Catalogs (`application/ai-catalog+json`) use custom domain prefixes combined with the standard, registered `+json` suffix.
+- **AI Skill Bundles**: Packages are represented using custom prefixes combined with standard, registered syntax suffixes for archives/compression: `+zip` (registered under RFC 9559) and `+gzip` (registered under RFC 6839). Unregistered suffixes like `+tgz` are avoided to prevent registration failure.
+- **Generic Formats (Markdown)**: Standard Markdown is represented by the generic, globally registered `text/markdown` media type. Because using `text/markdown` in isolation does not indicate the document is an AI Skill (unlike custom prefixes), we use the standard IANA `profile` parameter (`profile=ai-skill`) to cleanly declare the semantic profile of the document while remaining 100% standard-compliant.
```
jonathanhefner:
```suggestion
- **Generic Formats (Markdown)**: Standard Markdown is represented by the generic, globally registered `text/markdown` media type. Because using `text/markdown` in isolation does not indicate the document is an Agent Skill (unlike custom prefixes), we use the standard IANA `profile` parameter (`profile=agent-skill`) to cleanly declare the semantic profile of the document while remaining 100% standard-compliant.
```

specification/ai-catalog.md
```diff
@@ -197,29 +197,34 @@ It MUST contain the following members:
...
   - `application/a2a-agent-card+json` — an A2A Agent Card
-  - `application/mcp-server-card+json` — an MCP Server Card
-  - Any other media type defined by a protocol specification (e.g.,
-    `application/agentskill+zip` for skill definitions)
+  - `application/mcp-server+json` — an MCP Server Card
+  - `application/ai-skills+zip` — an AI Skill bundle (ZIP archive)
+  - `application/ai-skills+gzip` — an AI Skill bundle (gzipped tarball)
+  - `text/markdown; profile=ai-skill` — an AI Skill defined in a standard Markdown file
+
+  These values are designed to align with official IANA media type registration standards. Standard ecosystem types use registered structured syntax suffixes (`+json`, `+zip`, `+gzip`). For generic payload formats (like `text/markdown`), the standard `profile` parameter is used to specify the semantic schema (e.g., `profile=ai-skill`). For any new or custom types not listed here, it is up to the specific client implementation to handle them correctly.
```
jonathanhefner:
```suggestion
  These values are designed to align with official IANA media type registration standards. Standard ecosystem types use registered structured syntax suffixes (`+json`, `+zip`, `+gzip`). For generic payload formats (like `text/markdown`), the standard `profile` parameter is used to specify the semantic schema (e.g., `profile=agent-skill`). For any new or custom types not listed here, it is up to the specific client implementation to handle them correctly.
```

specification/ai-catalog.md
```diff
@@ -2154,13 +2146,13 @@ contains multiple artifact types:
...
         "displayName": "Example Plugin MCP Server",
-        "mediaType": "application/mcp-server-card+json",
+        "type": "application/mcp-server+json",
         "url": "https://github.com/anthropics/claude-plugins-official/blob/main/plugins/example-plugin/server-card.json"
       },
       {
         "identifier": "urn:claude-plugin:anthropic:example-plugin:skills",
         "displayName": "Example Plugin Skills",
-        "mediaType": "application/agentskill+zip",
+        "type": "application/ai-skills+zip",
```
jonathanhefner:
```suggestion
        "type": "application/agent-skills+zip",
```

adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,55 @@
...
+We will completely remove the `mediaType` field from the `Attestation` object schema under `trustManifest`. Instead of requiring an explicit `mediaType` inside the attestation entry, consumers can resolve or infer the format directly from the HTTP response headers when fetching the attestation `uri`, or from the file extension in the URL path.
+
+## Rationale
+
+### Suffixes and Parameters in `CatalogEntry`
+To ensure that our recommended types are fully registerable as official IANA media types, we adhere strictly to the IANA Structured Syntax Suffix registry rules (RFC 6838/6839):
+- **Custom Ecosystem Formats**: Types like A2A Agent Cards (`application/a2a-agent-card+json`), MCP Servers (`application/mcp-server+json`), and Nested Catalogs (`application/ai-catalog+json`) use custom domain prefixes combined with the standard, registered `+json` suffix.
+- **AI Skill Bundles**: Packages are represented using custom prefixes combined with standard, registered syntax suffixes for archives/compression: `+zip` (registered under RFC 9559) and `+gzip` (registered under RFC 6839). Unregistered suffixes like `+tgz` are avoided to prevent registration failure.
```
jonathanhefner:
```suggestion
- **Agent Skill Bundles**: Packages are represented using custom prefixes combined with standard, registered syntax suffixes for archives/compression: `+zip` (registered under RFC 9559) and `+gzip` (registered under RFC 6839). Unregistered suffixes like `+tgz` are avoided to prevent registration failure.
```

adr/0014-media-type-to-type.md
```diff
@@ -0,0 +1,55 @@
...
+We will rename the `mediaType` field to `type` in the `CatalogEntry` schema.
+The `type` field is an open text format, so any string value is accepted. To ensure clarity of ownership and governance, the recognized "known types" are partitioned into core protocol types and integrated third-party/ecosystem types:
+
+#### Core Protocol Types (Governed by the AI Catalog WG)
+- `application/ai-catalog+json` — nested AI Catalog
+- `application/agent-card+json` — reserved for a generic Agent Card format
+
+#### Integrated Ecosystem & Third-Party Types (Governed externally)
+- `application/a2a-agent-card+json` — A2A Agent Card
+- `application/mcp-server+json` — MCP Server Card
```
tadasant:
```suggestion
- `application/mcp-server-card+json` — MCP Server Card
```

specification/ai-catalog.md
mindpower: Changed back to agent-skills+zip now. Thanks @jonathanhefner

Wolfe-Jam:
hey guys, love the cards. not wishing to put yet another decision burden on your track — but I author a fair few Agent Skills (~40) and I'd love to see **tags carry discovery**. one word and you have everything, no matter how deep the catalog becomes — `skills`, `context`, `security`, images, whatever — a single tag finds them all, and it doesn't care how the entry's encoded.

I just tried the format out for myself (published a catalog with five entries) and that's where it landed for me. thanks again for the work on this.

— *Assisted by Claude (Opus 4.8) · Approved by James Wolfe (@Wolfe-Jam)*

tadasant (approved):

jonathanhefner (approved):

darrelmiller (approved):

dsp-ant (approved):

PR #19, opened 2026-03-05
spec: Add "AI Actor Naming Standard" specification

mindpower:
This PR introduces the "AI Actor Naming Standard," a foundational specification for ensuring interoperability, discovery within the AI agent ecosystem.

### Summary
As AI agents begin to interact across organizational boundaries, the lack of a common naming protocol creates ambiguity and hinders discovery. This standard solves this by defining a secure, multi-layer naming system that separates an agent's human-readable name from its stable, machine-readable logical address.

Core ### Concepts
The standard is built on two primary fields:

1. `displayName` (**Short Name**): A human-readable name used for discovery in UIs (e.g., "Google Tax Bot"). This name can be simple, intuitive, and is not required to be unique.

2. `identifier` (**Agent Identifier**): A globally unique, immutable logical address that follows a URN (Uniform Resource Name) convention. This acts as the permanent handle that developers use to reference the agent, ensuring stability even if underlying infrastructures change.

### The URN Specification (**identifier**)
The `identifier` provides a verifiable and hierarchical naming structure. It MUST follow the format: `urn:agent:<publisher-domain>:<namespace>:<agent-name>`

- This format establishes clear domain authority (<publisher-domain>) and allows for organized namespaces (e.g., productivity, finance).
- Example: `urn:agent:salesforce.com:productivity:slide-generator`

PR #39, opened 2026-06-08
spec: make Catalog Entry displayName optional (not required) (ADR 0016)

tadasant:
## What this proposes

Make `displayName` **OPTIONAL** on the **Catalog Entry** rather than required — i.e. move it from the required members of a Catalog Entry ([`ai-catalog.md` §Catalog Entry](https://github.com/Agent-Card/ai-catalog/blob/main/specification/ai-catalog.md#catalog-entry)) into the OPTIONAL members, drop it from the **Minimal Catalog** (Level 1) conformance requirement, and add guidance that it SHOULD be set **only when the referenced artifact does not already carry its own canonical human-readable name**.

The intent: when an entry points at an artifact that names itself (an A2A Agent Card, an MCP Server Card), the artifact is the single source of truth and the entry shouldn't duplicate the name. When an entry points at something opaque that has *no* embedded name (a raw dataset, a model blob, a skill bundle), the entry is the only place a human-readable name can live — so the field stays available for exactly that case.

This change deliberately does **not** touch `displayName` on the **HostInfo** or **Publisher** objects — those name the catalog host and the publishing org respectively, which is genuinely catalog-authored metadata with no other home.

### Alternative considered: removing the field entirely

My first instinct was to remove `displayName` from the entry level altogether and always defer to the artifact, but I'm not suggesting it because that would mean opaque artifacts have nowhere else to put a name. [`examples/ai-catalog.json`](https://github.com/Agent-Card/ai-catalog/blob/main/specification/examples/ai-catalog.json) has an `application/parquet` dataset entry; a Parquet file, a model blob, or a skill `.zip` doesn't embed a self-describing display name the way a card does. Removing the field outright would leave those artifacts nameless in a catalog. So "optional, with guidance" is the proposal; "remove entirely" is the rejected alternative.

## Why make it optional — the duplication / drift problem

A Catalog Entry already points at the full artifact via `url` or `data`. For the card-shaped artifact types this spec calls out by media type, the human-readable name is **already present in the artifact itself**:

- **A2A Agent Card** — `name` is a **REQUIRED**, human-readable field. From the A2A protobuf spec ([`specification/a2a.proto`](https://github.com/a2aproject/A2A/blob/main/specification/a2a.proto), `message AgentCard`):
  ```proto
  // A human readable name for the agent.
  // Example: "Recipe Agent"
  string name = 1 [(google.api.field_behavior) = REQUIRED];
  ```
- **MCP Server Card** — already carries a human-readable display name. In the [Server Card schema](https://github.com/modelcontextprotocol/experimental-ext-server-card/blob/main/schema.json), `name` is a reverse-DNS machine identifier (`^[a-zA-Z0-9.-]+/[a-zA-Z0-9._-]+$`), while **`title`** is documented as *"Optional human-readable title or display name for the MCP server"*, and `description` (required) is human-readable too.

So for the two flagship artifact types, *requiring* `displayName` on the entry means **every catalog publisher copies a value that already exists in the document the client is about to fetch**, and then has to keep the two in sync forever. When they drift — and over time they will — a consumer has two conflicting human-readable names for the same thing and no principled way to pick. Making the field optional (and recommending omission for self-naming artifacts) keeps the name fresh and authoritative at its source, while the entry stays a thin pointer.

## What's in the diff

- **Normative:** `displayName` moved from required → OPTIONAL members, with a SHOULD-guidance paragraph (set it only when the artifact has no canonical name; omit it for self-naming cards; the artifact's own name wins on conflict). Removed from the Minimal Catalog (Level 1) required-at-minimum list.
- **Examples (all of them, for internal consistency):** card-backed entries (A2A cards, MCP Server Cards, registry `server.json`, plugin manifests) now omit `displayName`; opaque entries (the `application/parquet` dataset, `application/agentskill+zip` skill bundles) and nested `application/ai-catalog+json` entries — which have no embedded canonical name — keep it. `HostInfo`/`Publisher` `displayName` untouched. The field still exists, so the class diagram and the "logical format" vocabulary lists are unchanged.
- **Conceptual-mapping tables** (MCP registry, Claude plugins) updated so the human-readable name maps to the artifact, with entry `displayName` framed as the fallback for nameless artifacts.

## Some history (for transparency)

As far as I can reconstruct from the public record, `displayName` was **required from its first appearance** rather than as the outcome of an explicit "should this be mandatory?" debate:

- It entered the draft as `name` in commit [`0271c14`](https://github.com/Agent-Card/ai-catalog/commit/0271c14873) (2026-03-31), already under "It MUST contain the following members."
- It was renamed `name` → `displayName` in [`1e8c70c9`](https://github.com/Agent-Card/ai-catalog/commit/1e8c70c9f3611135ca5fb7002781a1dca19c8e89) per the AI Actor Naming Standard ([PR #19](https://github.com/Agent-Card/ai-catalog/pull/19)), still required, and shipped via [PR #27](https://github.com/Agent-Card/ai-catalog/pull/27).

I went looking for a written rationale for the *mandatory* status specifically — across the ADRs (0001–0013), issues, and discussions — and couldn't find one. The closest articulations of the field's *purpose* are [PR #19](https://github.com/Agent-Card/ai-catalog/pull/19) (*"a human-readable name used for discovery in UIs … not required to be unique"*) and [ADR 0013](https://github.com/Agent-Card/ai-catalog/blob/main/adr/0013-authoring-vs-distribution-formats.md) framing the catalog as *"the authoring and discovery format — a human-readable JSON document."* I might simply be missing the discussion where requiring it was decided — **if there's prior context I've overlooked, please point me to it**, since it likely bears directly on the tradeoff.

## Counter-argument to landing this

The strongest argument I can construct *for keeping it required* is **browsing UX**: a UI that renders a list of N artifacts shouldn't have to fetch N separate cards just to show names, so requiring `displayName` on the entry makes the catalog self-describing for a cheap list view.

However, that's a steep trade-off for a permanent duplication/sync cost; one we'd like to avoid making on the MCP side. I'm not sure I'm convinced that UIs of AI Catalogs will be a major use case of AI Catalog (please correct me if you foresee otherwise, would love to understand the use case), and the optional form still lets a publisher populate `displayName` on every entry if they want the self-describing list view. It just doesn't force the card-backed entries to carry a redundant copy.

If there's a use case beyond browsing that makes the field load-bearing enough to mandate, would love to understand.

### DIFF
--- a/adr/0016-displayname-optional.md
+++ b/adr/0016-displayname-optional.md
@@ -23,7 +23,7 @@ Not all artifacts self-name, however. Opaque artifacts referenced by a catalog �
 
 Guidance: a publisher SHOULD set `displayName` only when the referenced artifact does not already carry its own canonical human-readable name. When the artifact does carry one, that artifact is the authoritative source and `displayName` SHOULD be omitted. When `displayName` *is* present, however, it takes precedence and is authoritative for display: a consumer SHOULD render it as given, even if it differs from a name carried by the referenced artifact — setting `displayName` is how a publisher deliberately overrides the artifact's own name.
 
-For an entry without `displayName`, a consumer SHOULD resolve a name in order: (1) the entry's `displayName` if present, (2) the referenced artifact's own canonical name (e.g. an A2A Agent Card `name` or MCP Server Card `title`) if already fetched or cached, then (3) the trailing segment of the entry's `identifier` URN. See [Resolving an Artifact's Display Name](../specification/ai-catalog.md#resolving-an-artifacts-display-name) in the specification.
+When rendering an entry, a consumer SHOULD resolve a name in order: (1) the entry's `displayName` if present, (2) the referenced artifact's own canonical name (e.g. an A2A Agent Card `name` or MCP Server Card `title`) if already fetched or cached, then (3) the trailing segment of the entry's `identifier` URN. See [Resolving an Artifact's Display Name](../specification/ai-catalog.md#resolving-an-artifacts-display-name) in the specification.
 
 This decision concerns only the Catalog Entry. `displayName` on `HostInfo` and `Publisher` is unchanged — those name the catalog host and publishing organization, which is genuinely catalog-authored metadata with no other home.
 
--- a/adr/0016-displayname-optional.md
+++ b/adr/0016-displayname-optional.md
@@ -1,0 +1,43 @@
+# ADR-0016: displayName Is Optional on a Catalog Entry
+
+## Status
+Proposed
+
+## Date
+2026-06-18 (Proposed)
+
+## Context
+A Catalog Entry is a thin pointer: it carries an `identifier`, a `mediaType`, and exactly one of `url` or `data` locating the full artifact. `displayName` (a human-readable name for the artifact) is a REQUIRED member of every Catalog Entry, and is listed among the required-at-minimum members of a Minimal Catalog (Level 1).
+
+For the card-shaped artifact types this spec calls out by media type, a human-readable name already lives inside the referenced artifact:
+
+- An **A2A Agent Card** has a REQUIRED, human-readable `name`.
+- An **MCP Server Card** carries a human-readable `title` (and a required human-readable `description`); its `name` is a reverse-DNS machine identifier.
+
+Requiring `displayName` on the entry therefore forces every catalog publisher to copy a value that already exists in the document the client is about to fetch, and to keep the two in sync indefinitely. When they drift, a consumer is left with two conflicting human-readable names and no principled way to pick.
+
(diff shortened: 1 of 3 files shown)


darrelmiller (approved):

darrelmiller:
My perspective on this is that with the identifier and description property, a registry can synthesize a friendly display name for their registry.  Picking the last segment of the URN identifier would likely be a reasonable substitute

mindpower:
Agree that for browsing/indexing, the `displayName` is useful to avoid the traffic to the A2A/MCP service, but it doesn't need to be a mandatory field. Making it optional makes perfect sense and should work well.

mindpower (approved):

PR #47, opened 2026-06-21
feat(spec): Trust Manifest threat modeling review - suggested updates.

### LINKED ISSUE #48: Review Trust Manifest Proposal
The trust manifest is a sensitive data model that requires security analysis.

- threat model
- ability to map to existing registries
- validation of MUST/SHOULD/MAY on manifest's fields.

muscariello:
- Require a present Trust Manifest to carry a substantive member (signature, attestations, provenance, or trustSchema); ADR-0015.
- Add a substrate-neutral distribution-mapping contract with OCI and xRegistry bindings; ADR-0014.

### DIFF
--- a/adr/0014-trust-manifest-artifact-binding.md
+++ b/adr/0014-trust-manifest-artifact-binding.md
@@ -1,0 +1,75 @@
+# ADR-0014: Bind Trust Manifest Signatures to Artifact Content
+
+**Status:** Accepted
+
+**Date:** 2026-06-20
+
+**Participants:** Luca Muscariello (Cisco)
+
+## Context
+
+[ADR-0009](0009-trust-manifest-substitution.md) acknowledged that Trust
+Manifest substitution is a real threat and called for a security review.
+That review was performed as a STRIDE threat model (see
+[trust-manifest-threat-model.md](../specification/trust-manifest-threat-model.md))
+and confirmed the central weakness: the `signature` covered only the
+Trust Manifest JSON, while the artifact reference (`url`, `data`,
+`mediaType`) lived on the Catalog Entry *outside* the signed payload.
+The only binding was `identity == identifier`. An attacker who controlled
+the catalog document could keep a validly-signed manifest and repoint
+`url` to a malicious artifact — the signature would still verify
+(threat T1 / finding F1).
+
+The threat model also compared the design against the Sigstore
+architecture from which it derives. Sigstore's Cosign signs the artifact
+*digest*, Fulcio binds identity through a CA, and Rekor witnesses the
+event in a transparency log. The Trust Manifest natively reproduced none
+of these; most importantly it lacked Cosign's digest binding.
+
+## Decision
+
+Bind a signed Trust Manifest to the artifact it describes and harden the
+verification story:
+
+1. **Subject binding.** A signed Trust Manifest MUST include a `subject`
+   object `{ url?, mediaType, digest }` committing to the artifact's
+   media type and content digest. `subject` is part of the signed
+   payload, so the artifact reference cannot be changed without
+   invalidating the signature. (Closes F1; mirrors Cosign.)
+2. **Signatures required at Level 3.** A Trusted Catalog MUST carry a
+   `signature`, `subject`, and `issuedAt` on each relied-upon entry, and
+   consumers MUST verify them. (Closes F2.)
+3. **Trust anchoring.** Verification MUST anchor `identity` to an
+   out-of-band trust root; a verified signature alone proves internal
+   consistency, not publisher authenticity. (Closes F3; substitutes for
+   Fulcio's CA-attested identity.)
+4. **Algorithm allowlist.** Consumers MUST reject `alg: none` and
+   symmetric algorithms and MUST use asymmetric JWS algorithms. (F4)
+5. **Freshness.** Add `issuedAt` (REQUIRED when signed) and `expiresAt`,
+   with anti-rollback and revocation guidance. (F5)
+6. **Catalog-level integrity.** Add an OPTIONAL top-level catalog
+   `signature` and recommend content-addressed (OCI) distribution. (F6)
+7. **Safe fetching.** Verification MUST defend against SSRF and
+   oversized responses. (F7)
+8. **Provenance-statement and publisher verification** procedures are
+   defined; publisher fields are advisory unless bound. (F8, F9)
+
+## Sigstore convergence
+
+The Trust Manifest remains a *format that carries* trust evidence, not a
+replacement for Sigstore services. The specification SHOULD allow
+Sigstore evidence to be first-class: a `sigstore-bundle` attestation
+(certificate + signature + Rekor inclusion proof), anchoring via the
+Sigstore TUF root, and a preference for transparency-log inclusion and
+keyless signing over long-lived publisher keys. This lets
+trust-sensitive deployments inherit Sigstore's full chain instead of a
+weaker re-implementation. Tracked as future work.
+
+## Consequences
+
+- The substitution attack from ADR-0009 is closed for signed manifests.
+- Level 3 conformance is stricter: presence of a manifest is no longer
+  sufficient; it must be signed and bound.
+- Residual gaps versus Sigstore remain (no native transparency log,
+  long-lived keys, trust-anchor bootstrapping) and are documented as
+  residual risks plus the Sigstore-convergence roadmap.
--- a/adr/0015-substantive-trust-manifest.md
+++ b/adr/0015-substantive-trust-manifest.md
@@ -1,0 +1,63 @@
+# ADR-0015: Require a Substantive Trust Manifest
+
+**Status:** Accepted
+
+**Date:** 2026-06-21
+
+**Participants:** Luca Muscariello (Cisco)
+
+## Context
+
+The Trust Manifest is an OPTIONAL companion to a Catalog Entry. Within
+it, however, only `identity` was unconditionally REQUIRED, and `identity`
+MUST equal the entry's `identifier`. Every other member — `signature`,
+`attestations`, `provenance`, `trustSchema`, and the informational fields
+— was independently OPTIONAL.
+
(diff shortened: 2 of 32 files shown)


muscariello:
@darrelmiller @mindpower I went through a threat modeling session on the trust manifest. This is the set of modifications that I suggest to address some of the comments by Pamela and others.

I did not do an accurate analysis since quite some time and if we want to get to v1 this is necessary.

ramizpolic (commented):
This PR can also resolve the issue (_Support multiple identities per catalog entry_ #52) depending if we have a decision. It may be useful extend this PR to include multiple trustmanifests, which would address both the concerns around identity and threat modelling in one go.

olijboyd:
I've implemented what this PR specifies end to end, because I wanted TomeVault's own catalog signed this way rather than waiting for v1, so I can offer some evidence from the implementing side. The subject binding closes the substitution case cleanly. Our vector suite runs the swap both ways, a validly signed pre-binding manifest accepts a substituted artifact, and the same swap fails once `subject` is inside the signed payload, which is ADR-0015's argument demonstrated in running code. The signatures verify under jwcrypto and authlib as well as our own verifier, so the signing input is interoperable rather than just self-consistent, and it isn't only a lab exercise either: tomevault.io serves its catalog with this binding today, subject digest bound to the mcp.json bytes actually being served, if anyone wants a live implementation to poke at.

Three things bit me on the way through, offered as notes rather than review findings.

The one that genuinely blocks a conformant implementation is a field-name mismatch the binding check depends on. The Catalog Entry prose defines the artifact's media type as `type`, "the media type declared in the `type` field", while `specification/examples/ai-catalog.json` and ADR-0014 both say `mediaType`, so `subject.mediaType` is checked against an entry field the normative prose never defines. We went with `mediaType` to match the examples and the binding, but it wants settling one way or the other before v1.

F7 can be satisfied by a fetch that still fails two ways, and I know because ours did: a rebind between the address check and the connection, and a redirect off a validated URL into somewhere private. Resolving once, pinning the connection to the validated address, and refusing redirects closed both, and it may be worth the ADR saying that much, because "validate the target" reads as satisfied by a one-time check in front of an ordinary HTTP client.

The large-integer caveat is real in practice too. An integer outside the JCS-safe range throws mid-canonicalisation, so a verifier has to catch that and return a rejection instead of letting it surface as an internal error.

If a reference verifier, or the threat-model document ADR-0014 points at (trust-manifest-threat-model.md isn't in the repo yet), would help this land, I'm glad to contribute either. The reason we lean on this PR so hard is that the attestations we attach are quality and consistency claims about content that drifts, and that kind of claim only means anything bound to the exact bytes it was made about. Once this lands I'd like to bring an attestation type for measured quality and consistency, but that's a follow-up and this PR comes first.

olijboyd:
AIR is letting a publisher sign their resource and a consumer record what it fetched. But there's no independent third party verification yet… Whether an instruction or skills file is any good, and whether it has quietly drifted since it was listed, is the part that actually protects the user, and today it can only live in a vendor extension. A grade sitting in a vendor extension is just a value. Anyone re-serving the catalog can flip it from bad to good and nothing in AIR tells the consumer to check.

I already setup TomeVault to grade AI instruction & skills files; now I've implemented AIR so it also signs the verdict, and monitors them for conformance and drift. The flat tv: values on the entry credit the original author and are fine for discovery, but they're only a hint, not something to trust. The grade you trust comes from a signed record which the entry references by digest. A consumer reads the grade from that signed record and ignores the flat values when the two disagree. I've signed 180k+ skills this way.

This is already built and public. The reference verifier is Apache-2.0 (https://github.com/tomevault-io/ai-catalog-reference), ships the substitution-attack tests, and verifies the record against the published key. A matching JavaScript verifier on npm checks the same records. You can pull one and check it yourself. Fuller write-up: https://tomevault.io/docs/attestation/agent-conformance-air-extension.

I'd like to help make third-party verification a real part of the standard so I'd value your read on whether this is the right way to close it.

adr/0014-trust-manifest-artifact-binding.md
```diff
@@ -0,0 +1,75 @@
+# ADR-0014: Bind Trust Manifest Signatures to Artifact Content
```
mindpower: The latest adr is 0016. Update the number here?

adr/0015-substantive-trust-manifest.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Date:** 2026-06-21
+
+**Participants:** Luca Muscariello (Cisco)
+
+## Context
+
+The Trust Manifest is an OPTIONAL companion to a Catalog Entry. Within
+it, however, only `identity` was unconditionally REQUIRED, and `identity`
+MUST equal the entry's `identifier`. Every other member — `signature`,
```
mindpower: I don't remember that we require that. 

The AI Catalog specification requires **alignment** (domain binding), not an exact match. 

The `identifier` and the `identity` will almost never be exact matches because they serve different purposes and use different URI schemes (e.g., a `urn:air:...` versus a `spiffe://...` or `did:web:...`). 

What the specification strictly requires is that the **authority/trust domain** embedded in the `identity` matches the **publisher domain segment** inside the `identifier`. 

For example,:
* **`identifier`**: `urn:air:example.com:travel:concierge` (Publisher domain = `example.com`)
* **`identity`**: `spiffe://example.com/travel/concierge` (Authority domain = `example.com`)

This satisfies the spec because the domains **align** (`example.com`), cryptographically binding the workload to the authorized publisher namespace. The strings themselves do not need to be exactly the same.

adr/0015-substantive-trust-manifest.md
```diff
@@ -0,0 +1,63 @@
...
+`identity`, a Trust Manifest MUST contain at least one *substantive* trust
+member:
+
+- a `signature` (with its required `subject` and `issuedAt`),
+- a non-empty `attestations` array,
+- a non-empty `provenance` array, or
+- a `trustSchema`.
+
+`identity` and `identityType` (which restate or describe the entry
+identifier) and the informational members `privacyPolicyUrl`,
```
mindpower: This statement is not correct.

`identity` and `identityType` do **not** restate or describe the entry `identifier`. 

The purpose if `identity` is entirely different from the `identifier`:
* **`identifier`** describes the *logical name* used for catalog search and discovery (e.g., `urn:air:example.com:travel:concierge`).
* **`identity` and `identityType`** describe the **Cryptographic Workload Principal**—meaning they define *who* actually owns and runs the agent at runtime (e.g., a `spiffe` type identity like `spiffe://example.com/travel/concierge`). 

Rather than **restating** the discovery name, `identity` and `identityType` provide the cryptographic anchor used for zero-trust security, runtime authentication, and verifiable compliance.

specification/examples/ai-catalog.json
```diff
@@ -15,7 +15,31 @@
...
       ],
       "url": "https://api.acme-corp.com/agents/acme-finance-agent.json",
-      "updatedAt": "2026-02-22T16:30:00Z"
+      "updatedAt": "2026-02-22T16:30:00Z",
+      "publisher": {
+        "identifier": "did:web:acme-corp.com",
+        "displayName": "Acme Services Inc."
+      },
+      "trustManifest": {
+        "identity": "urn:example:agent-finance-001",
```
mindpower: 1. this is not an valid agent identifier name (should start with urn:air:<domain name>:...)
2. The identityType below is "did" but the identity is not did.

adr/0015-substantive-trust-manifest.md
```diff
@@ -0,0 +1,63 @@
+# ADR-0015: Require a Substantive Trust Manifest
```
mindpower: Please update the sequence number.

adr/0015-substantive-trust-manifest.md
muscariello: @mindpower fixed.

adr/0015-substantive-trust-manifest.md
muscariello: @mindpower correct. And fixed. Thanks.

adr/0014-trust-manifest-artifact-binding.md
muscariello: fixed - this counter keeps increasing. Once this is merged we're ok.

specification/examples/ai-catalog.json
muscariello: done

muscariello:
A @ramizpolic  is including multi-attestation in here  https://github.com/Agent-Card/ai-catalog/pull/91 I will not propose anything in this PR. But it would be important to review the multi-attestation proposal too. @mindpower

specification/ai-catalog.md
```diff
@@ -493,10 +552,57 @@ provenance:
...
+The Subject object binds a Trust Manifest to the specific artifact it
+describes, closing the substitution gap in which an attacker who
+controls the catalog document leaves a validly-signed Trust Manifest in
+place but repoints the entry to a different artifact. Because the
+Subject is part of the signed payload, the artifact reference and its
+content digest cannot be changed without invalidating the signature.
+
+A Subject object MUST contain:
+
+`mediaType`
```
tadasant:
```suggestion
`type`
```

Other instances as well

adr/0017-trust-manifest-artifact-binding.md
```diff
@@ -0,0 +1,75 @@
+# ADR-0017: Bind Trust Manifest Signatures to Artifact Content
```
tadasant: Will want to bump these ADR numbers to be 19+

tadasant (commented):
These changes are good with me 👍

muscariello:
@darrelmiller @mindpower @tadasant @dsp-ant 
I implemented all the requested changes. 

I think we can merge this one now and move one.

mindpower:
@muscariello ,Thanks for addressing my previous comments. The changes look great. 

Huge +1 for adding the formal threat model review documents.

mindpower (approved):

PR #54, opened 2026-06-26
docs: add server.json as a first-class catalog type (application/mcp-server+json)

tadasant:
> ⚠️ **AI-assisted draft.** This PR's content (spec prose and description) was drafted by an AI assistant working on Tadas's behalf. Tadas will review, vet, and rewrite as needed, and will remove this disclosure once he has done so. Until then, treat this as a steered placeholder, not Tadas's finished position.

## Stacked on #53

This PR is **stacked on top of #53** (`docs/update-server-json-mapping`). Its base branch is `docs/update-server-json-mapping`, not `main` — review/merge #53 first. The diff against its base is just the `server.json`-as-first-class change below.

## What this does

Reintroduces the MCP Registry `server.json` mapping that #53 removed, but as a **first-class cataloged artifact type** rather than an experimental footnote.

### Key changes

- **New known media type `application/mcp-server+json`** for an MCP Registry `server.json` document, registered in both the spec's known-types list and [ADR 0014](https://github.com/Agent-Card/ai-catalog/blob/main/adr/0014-media-type-to-type.md) (under "Integrated Ecosystem & Third-Party Types").
- **New appendix `Mapping to MCP Registry server.json`** describing the first-class mapping: a Catalog Entry whose `url` points to the `server.json` document with `type` `application/mcp-server+json`. Includes a conceptual mapping table, an entry example, an "MCP Registry as AI Catalog" example, and a "Relationship to MCP Server Cards" section.
- **Two-types framing.** The appendix is explicit that the MCP ecosystem has two distinct server artifacts, each with its own `type`:
  - **MCP Server Card** → `application/mcp-server-card+json` (static discovery / connection details)
  - **MCP Registry `server.json`** → `application/mcp-server+json` (installable package descriptor)

  A catalog entry references whichever fits; a domain MAY list both.

- **Examples use real MCP Registry URLs.** Every `server.json` example `url` uses the actual registry API form `https://registry.modelcontextprotocol.io/v0/servers/{url-encoded-name}/versions/{version}` (the `/` in the reverse-DNS name encoded as `%2F`), with verified-real servers/versions (`com.pulsemcp/remote-filesystem` 0.1.5, `com.pulsemcp.servers/pulse-fetch` 0.2.14). The appendix also explains how a registry name maps to the `urn:air:{publisher}:{namespace}:{name}` identifier per [ADR 0015](https://github.com/Agent-Card/ai-catalog/blob/main/adr/0015-agent-identifier-naming.md).

- **Omitted duplicative `version`/`description` from the example entries.** Mirrors the same change in #53's Server Card examples: these restate values the `server.json` document already carries (the drift argument that keeps `displayName` out of an entry). The example entries drop `version`/`description`; the conceptual mapping table keeps documenting them but its `description`/`version` rows justify the omission, parallel to the `title` row. For `server.json`, the entry's version-pinned `url` already identifies which version it points at.
- **Applied the same drift-avoidance treatment to `repository`.** `server.json` carries its own `repository`, so mapping it into entry `metadata.repository` duplicated a drift-prone value. The example already demonstrated the redundancy — the repo URL also appears in `trustManifest.provenance[].sourceId` — so the `metadata.repository` block is dropped from the example and the table row now justifies the omission (source/provenance links belong in the Trust Manifest). This mirrors the Server Card `repository` fix inherited from #53.
- **Dropped `updatedAt` from the entry example (registry-timestamp drift).** The single-entry example carried `"updatedAt": "2026-03-15T10:00:00Z"`, but the MCP Registry response already exposes its own `publishedAt`/`updatedAt` under `_meta` (verified live: `registry.modelcontextprotocol.io` returns `_meta["io.modelcontextprotocol.registry/official"].updatedAt`), so the entry field duplicated a value the source carries — and the hardcoded date had already drifted from the real one. Removed it from the example (bringing it to structural parity with #53's Server Card entry example, which carries no `updatedAt`) and extended the `_meta` mapping-table row to record that the registry timestamps stay in the artifact.
- **`version` keeps its multi-version carve-out here (deliberate divergence from the Server Card side).** #53 dropped the multi-version carve-out from the Server Card `version` row because a remote MCP server serves exactly one card. On the Registry `server.json` side that case is real — a registry genuinely serves multiple versions of a name — so this appendix's `version` row retains the "omitted unless the catalog lists multiple versions of one identifier" language, plus the note that the version-pinned `url` already identifies which version an entry points at.

### Note on the naming distinction (issue #9 / PR #18)

ext-server-card issue #9 / PR #18 deliberately chose `application/mcp-server-card+json` for the **Server Card** so it would *not* collide with the Registry `server.json` concept. This PR is the other half of that decision: it allocates the unqualified `application/mcp-server+json` to the Registry `server.json` document itself. The two types are complementary, not competing.

## Verification

- [x] **Spec builds locally** with the exact CI command (`python tools/build_spec.py`), exit 0, `dist/index.html` produced.
- [x] **New type registered in both places** — `application/mcp-server+json` appears in the spec known-types list and ADR 0014 (verified the HTML renders it; 11 occurrences in built output across the new appendix + lists).
- [x] **Cross-reference resolves** — the appendix's link to `#mapping-to-mcp-servers` matches the generated heading `id` (verified in built HTML).
- [x] **No new broken anchors** — built-HTML link/`id` audit shows the only unresolved anchors (`metadata-extensibility`, `resolving-an-artifacts-display-name`, `version-handling`) are **pre-existing on `main`**; none are introduced or referenced by this change.
- [x] **All JSON examples valid** and use `type: "application/mcp-server+json"` with `urn:air:` identifiers.
- [x] **Duplicative fields removed cleanly** — no `server.json` example entry carries `version`/`description`/`metadata.repository`/`updatedAt` (grep confirmed the appendix's only remaining `updatedAt` is the explanatory `_meta` table row, not an example value); version remains recoverable from the version-pinned `url`, the repo from `trustManifest.provenance`, and the timestamps from the registry `_meta`; the mapping table keeps documenting `version`/`description`/`repository`/`_meta`, each with its omission rationale. All 19 JSON examples still parse.
- [x] **Registry-timestamp drift verified live** — `curl` of `registry.modelcontextprotocol.io/v0/servers/com.pulsemcp%2Fremote-filesystem/versions/0.1.5` returns `_meta["io.modelcontextprotocol.registry/official"]` with `publishedAt`/`updatedAt` (`2026-06-26…`), confirming the removed entry `updatedAt` was both duplicative and stale against the source.
- [x] **Example URLs verified against the live registry** — both example URLs (`com.pulsemcp%2Fremote-filesystem/versions/0.1.5`, `com.pulsemcp.servers%2Fpulse-fetch/versions/0.2.14`) return HTTP 200 from `registry.modelcontextprotocol.io`, with response `server.name`/`version`/`repository` matching the example values; the prose's `/v0/servers/{name}/versions/{version}` + `server`/`_meta` envelope description was confirmed against that response.
- [x] **Inherits #53's discovery-flow reframe** — the shared Server Card appendix's "Decentralized Discovery" section (reframed per [ext-server-card PR #36](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/36)) merges in cleanly; build passes on this branch.
- [x] **Fresh-eyes review** performed in-process on the diff vs. the base branch; feedback addressed.
- [x] **CI green** on the latest commit (confirmed via wait-for-CI after the final push).

## Status

**Draft — do not merge.** Stacked on #53; awaiting Tadas's review and rewrite.

### DIFF
--- a/adr/0014-media-type-to-type.md
+++ b/adr/0014-media-type-to-type.md
@@ -29,6 +29,7 @@ The `type` field is an open text format, so any string value is accepted. To ens
 #### Integrated Ecosystem & Third-Party Types (Governed externally)
 - `application/a2a-agent-card+json` — A2A Agent Card
 - `application/mcp-server-card+json` — MCP Server Card
+- `application/mcp-server+json` — MCP Registry `server.json` document
 - `application/agent-skills+json` — Agent Skill Metadata json file
 - `application/agent-skills+md` — Agent Skill defined in a standard Markdown file (the suffix +md is to be registered)
 - `application/agent-skills+zip` — Agent Skill bundle in a ZIP archive
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -218,6 +218,7 @@ It MUST contain the following members:
 
     - `application/a2a-agent-card+json` — an A2A Agent Card
     - `application/mcp-server-card+json` — an MCP Server Card
+    - `application/mcp-server+json` — an MCP Registry `server.json` document
     - `application/agent-skills+json` — Agent Skill Metadata json file
     - `application/agent-skills+md` — an Agent Skill defined in a standard Markdown file (the suffix `+md` is to be registered)
     - `application/agent-skills+zip` — an Agent Skill bundle (ZIP archive)
@@ -1754,6 +1755,12 @@ to reference each server's **MCP Server Card**
 ([SEP-2127](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127))
 as the artifact content of a Catalog Entry.
 
+A Server Card is a static discovery document for a single HTTP-based MCP
+server (its identity and connection details). It is distinct from the MCP
+Registry's `server.json`, an installable-package descriptor — a separate
+artifact type mapped in
+[Mapping to MCP Registry server.json](#mapping-to-mcp-registry-server-json).
+
 ## Overview
 
 An MCP Server Card is a static discovery document for an individual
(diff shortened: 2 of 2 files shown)


darrelmiller:
@tadasant  Now that #53  is merged, can this come out of draft?

tadasant:
@darrelmiller a hesitation I have with officially landing this is that I want to make sure we are clear on the use case for supporting MCP Registry entries here, specifically with relation to the [MCP Registry spec](https://github.com/modelcontextprotocol/registry) itself.

We are clear on the need for Server Cards (which correspond to remote servers), and [the discovery mechanism](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/36) we (MCP) are pushing as the core use case makes sense. I'm less clear on whether endorsing `application/mcp-server+json` within AI Catalog necessarily supersedes the MCP Registry spec. i.e. would landing these be encouraging everyone who has built an MCP Registry to migrate over to AI Catalog? That would be a major shift in official position I think we (specifically MCP, not AI Catalog's problem) should take time to collect feedback on.

There is, of course, nothing stopping someone from using `application/mcp-server+json` in an AI Catalog implementation regardless of whether we land this example or not, so I'm tempted to leave this be and see if the ecosystem builds traction for this way of building a registry over the MCP Registry model (which does already have pretty widespread traction and utility). Maybe we can collect use cases and carve out an outcome like, "when you should use the MCP Registry" vs. "when you should use the AI Catalog" to catalog your server.json values.

Do you know of anyone personally keen to adopt server.json values in AI Catalog? Would love to have a chat with them to understand the details.

tadasant:
Where this stands as of Aug 20, 2026 meeting - I'd like to hold this example in draft until I've had some conversations with folks who are running server.json's / original MCP Registries in production, and how they feel about what value the more heterogenous AI Catalog format adds to them (would they migrate? will they keep using the MCP-only Registry?). This should help inform whether MCP wants to endorse using server.json values inside AI Catalog, or keep the integration point limited to Server Cards (as we are currently pursuing [here](https://github.com/modelcontextprotocol/experimental-ext-server-card#what-is-a-server-card)).

Please reach out or post here if you are such a persona and have an opinion on this.

Wolfe-Jam:
Hey Tadas — you mentioned wanting input from folks running server.json's / MCP Registries in production. Here's mine: 5 on `one.faf/*` (rust-faf-mcp, gemini-faf, claude-faf, grok-faf, faf-context (ide/cursor)).

Yes, I'd use this. The two-types split (Server Card vs `server.json`) is the right call — they genuinely describe different lifecycle stages, and I wouldn't want to lose either.

One friction point, from actually living it this week — you're probably already tracking this: the `publisher`/`identifier` mapping is clean for domain-verified namespaces, but I don't see a story yet for `io.github.*` publishers — still the majority of the registry. No domain to anchor a `did:web` to.

Concrete case, live right now: `rust-faf-mcp` exists under both `one.faf/rust-faf-mcp` (current) and `io.github.Wolfe-Jam/rust-faf-mcp` (frozen, several versions behind) — same package, two identities, nothing linking them (filed as modelcontextprotocol/registry#1556). Not a blocker on adopting this — just the one real gap between "I'd migrate" and "here's what migrating actually requires."

Happy to be a live test case if useful.

— *Assisted by Claude · Approved by James Wolfe*

Wolfe-Jam:
Update since my last comment — the catalog's live and current: https://faf.one/.well-known/ai-catalog.json (8 entries, last one added today). Real heterogeneity: `.faf`/`.fafm` (IANA-registered, outside this WG's own type list), two Server Cards, two agent-card flavors. Not attached either way — happy to tweak if useful as a reference while you're validating the format.

PR #60, opened 2026-07-02
spec: bring description/version to displayName's authoritative-source parity (ADR-0018)

tadasant:
## What & why

ADR-0016 made the Catalog Entry `displayName` OPTIONAL and gave it a precise **authoritative-source** rule: omit it when the referenced artifact already carries its own canonical name (A2A Agent Card `name`, MCP Server Card `title`) to avoid drift; when present it takes precedence for display; and a companion **"Resolving an Artifact's Display Name"** section defines the consumer resolution order.

`displayName` is not the only entry member that restates a value the referenced artifact already carries. `description` and `version` have the **same duplication-and-drift problem against the same artifact**, but the `description`/`version` **field definitions** were still bare one-liners with no equivalent guidance. That's incoherent — a publisher reading the field list is told to omit `displayName` against the Server Card's `title` but given no steer on `description`/`version`, which map onto the Server Card's own `description` and `version`. This PR closes that gap at the field-definition level.

Field overlap this PR is built on:

| Entry field | A2A Agent Card | MCP Server Card / SEP-2127 | MCP Registry `server.json` |
|---|---|---|---|
| `displayName` | `name` (required) | `title` (optional) | `title` |
| `description` | `description` (required) | `description` (optional) | `description` |
| `version` | `version` (required) | `version` (required) | `version` (required) |

## Changes

- **New `adr/0018-entry-field-authoritative-source-parity.md`** (Status: Proposed) — extends ADR-0016's rule to `description` and `version`, with an explicit in/out **Scope** section (out: `tags`, `updatedAt`, `publisher`/`trustManifest`, `metadata`, `identifier`/`type`/`url`/`data`, each with reasoning).
- **`specification/ai-catalog.md`:**
  - `description` **field definition** — parallel authoritative-source guidance + pointer to a new resolution section.
  - `version` **field definition** — authoritative-source guidance **adapted for its structural role**: it's part of the `identifier`+`version` uniqueness key and consumers sort on it, so it stays REQUIRED for multi-version listings and a present value is sort-authoritative (SHOULD equal the artifact's version; a contradiction is a publishing error, not a free-form override) rather than a cosmetic override.
  - New **"Resolving an Artifact's Description"** section next to "Resolving an Artifact's Display Name".
  - Claude Plugins mapping-appendix `description` row aligned with the existing plugin-`name`/`displayName` row.

No schema change — `description` and `version` are already OPTIONAL in the CDDL. This is guidance only and is **not** a breaking change (existing catalogs that populate these fields remain conformant).

## Scope

Deliberately tight: one ADR + two field-definition edits + one resolution section + one mapping-row alignment. No refactoring of unrelated spec parts.


darrelmiller:
Took me a couple of times to read this but now I grok it, I like this change a lot.  Thank you.

darrelmiller (approved):

Tehsmash (approved):

PR #73, opened 2026-07-08
fix: clean up mediaType-to-type rename fallout (ADR 0014/0015)

### LINKED ISSUE #66: ADR 0014 rename fallout: duplicate 'type' in Attestation schema, invalid OCI example, stale example file
The `mediaType` -> `type` rename (ADR 0014) appears to have been applied mechanically across the spec, which left three defects:

## 1. Duplicate `type` member in the Attestation schema

ADR 0014 removed `mediaType` from the Attestation object, but the CDDL now declares `type` twice - the second one looks like a leftover of the removed `mediaType` that was renamed instead of deleted:

```cddl
Attestation = {
  type: text,
  uri: text,
  type: text,        ; <- duplicate
  ? digest: text,
  ? size: uint,
  ? description: text
}
```

Duplicate map keys make the CDDL invalid. The mermaid diagram in "Data Model Overview" has the same duplicate (`type string` appears twice in `class Attestation`).

Location: `specification/ai-catalog.md`, "CDDL Schema > Trust Manifest" and "Data Model Overview".

## 2. OCI Image Index example uses `type` instead of `mediaType`

The example in "Mapping to OCI Distribution > OCI Image Index Example" uses `"type"` on the index and on each manifest descriptor. The OCI image spec requires the field name `mediaType` (e.g. `"mediaType": "application/vnd.oci.image.index.v1+json"`). The rename was applied to third-party OCI JSON that this spec does not control, so the example is not valid OCI.

## 3. `specification/examples/ai-catalog.json` was never migrated

The standalone example file still uses `mediaType`, so its entries are missing the required `type` member and the file does not validate against the spec's own CDDL. Its identifiers also still use the pre-ADR-0015 `urn:example:...` style instead of the standardized `urn:air:{publisher}:{namespace}:{name}` form ("All standard examples and specifications will be updated to follow this URN convention").

liady:
Fixes #66

The `mediaType` -> `type` rename (ADR 0014) was applied mechanically in a few places it should not have been, and skipped one place it should have reached. This PR cleans up all three:

## 1. Duplicate `type` member in the Attestation schema

ADR 0014 removed `mediaType` from the Attestation object, but it was renamed to `type` instead of deleted, leaving the CDDL with a duplicate map key (invalid CDDL). The mermaid diagram in "Data Model Overview" had the same duplicate. Both now declare `type` once.

## 2. OCI Image Index example restored to `mediaType`

The example in "Mapping to OCI Distribution > OCI Image Index Example" used `"type"` on the index and manifest descriptors. The OCI image spec requires the field name `mediaType`, so the example was not valid OCI. The rename does not apply to third-party OCI JSON; restored `mediaType` in the three descriptor positions (`artifactType` fields untouched).

## 3. `specification/examples/ai-catalog.json` migrated

The standalone example file still used `mediaType`, so its entries were missing the required `type` member and did not validate against the spec's own CDDL. Migrated to `type`, and updated the identifiers from the pre-ADR-0015 `urn:example:...` style to the standardized `urn:air:{publisher}:{namespace}:{name}` form (with the host identifier aligned to the `did:web:` style used in the spec's other examples).

## Verification

- `uv run --locked python tools/build_spec.py specification/ai-catalog.md ... --config specification/respec-config.json` builds cleanly
- `specification/examples/ai-catalog.json` parses as valid JSON and now conforms to the CatalogEntry CDDL (has `identifier`, `type`, and exactly one of `url`/`data`)

### DIFF
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -1152,7 +1152,6 @@ classDiagram
     class Attestation {
         type string
         uri string
-        type string
         digest string
     }
     class ProvenanceLink {
@@ -1324,7 +1323,6 @@ TrustSchema = {
 Attestation = {
   type: text,
   uri: text,
-  type: text,
   ? digest: text,
   ? size: uint,
   ? description: text
@@ -1668,11 +1666,11 @@ authored by hand:
 ```json
 {
   "schemaVersion": 2,
-  "type": "application/vnd.oci.image.index.v1+json",
+  "mediaType": "application/vnd.oci.image.index.v1+json",
   "artifactType": "application/ai-catalog+json",
   "manifests": [
     {
-      "type": "application/vnd.oci.image.manifest.v1+json",
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
       "digest": "sha256:aaa111...",
       "size": 1024,
       "artifactType": "application/a2a-agent-card+json",
@@ -1682,7 +1680,7 @@ authored by hand:
       }
     },
     {
-      "type": "application/vnd.oci.image.manifest.v1+json",
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
       "digest": "sha256:bbb222...",
       "size": 512,
       "artifactType": "application/mcp-server-card+json",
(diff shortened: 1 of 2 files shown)


tadasant (approved):
Thank you!

PR #74, opened 2026-07-08
fix: small editorial fixes across spec, README, ReSpec config, and docs

### LINKED ISSUE #67: Small editorial fixes across spec, README, ReSpec config, and docs
A collection of small, mechanical fixes across the spec, README, ReSpec config, and docs site. Grouped into one issue to avoid noise; none of them changes normative behavior.

1. **README typo** - `README.md` ("Static Discovery" section): `./well-known/ai-catalog.json` should be `/.well-known/ai-catalog.json`.

2. **Stale ReSpec abstract** - `specification/respec-config.json` still says the appendices describe a mapping to "the MCP Registry server.json format". The appendix was rewritten to map to MCP Server Cards (SEP-2127), so the published page's abstract misdescribes its own contents.

3. **Truncated sha256 digests in examples** - several examples use `"sha256:a1b2c3d4e5f6"` (12 hex chars), but the Digest Format section defines the value as the hex encoding of the hash output (64 chars for SHA-256). Affected: `specification/ai-catalog.md` (multi-artifact example, dual-protocol example, MCP mapping example) and `docs/examples/multi-protocol-agent.md`. Other examples already use full-length digests.

4. **Publisher domain mismatch in the MCP mapping example** - in "Mapping to MCP Servers", the example entry uses identifier `urn:air:example.com:mcp:finance-server` while its publisher is `did:web:acme-corp.com` and all URLs are on `acme-corp.com`. Per the URN rule, `{publisher}` is the publishing organization's domain, so this should be `urn:air:acme-corp.com:...` (the "MCP Servers as an AI Catalog" example a few lines below already does it correctly). The conceptual-mapping table repeats the `example.com` form.

5. **Well-Known URI prose implies Trust Manifests are fetched from `url`** - "the client SHOULD use the `url` entries to retrieve individual artifacts and their associated Trust Manifests". Trust Manifests are carried inline on catalog entries; they are not retrieved from the artifact URL.

6. **Docs conformance-level drift** - `docs/getting-started.md` says "You're already at Level 2 if you're serving the catalog at /.well-known/ai-catalog.json". Per the spec this is inverted: Level 2 requires the `host` object, and serving at the well-known URI is only a MAY.

7. **Docs trust-guide nits** - `docs/guides/adding-trust.md` shows `"identityType": "did"` on an identity that is a `urn:air:` URN, not a DID; and says "Three `relation` types are defined" where the spec defines `relation` as an open string with example values.

liady:
Fixes #67

Small mechanical fixes across the spec, README, ReSpec config, and docs site. No normative behavior changes.

- **README**: `./well-known/ai-catalog.json` -> `/.well-known/ai-catalog.json`
- **ReSpec abstract** (`specification/respec-config.json`): the appendix maps to MCP Server Cards, not "the MCP Registry server.json format"; the abstract now matches the appendix content
- **Truncated digests**: replaced 12-char `sha256:a1b2c3d4e5f6` example values with full 64-char hex digests, matching the Digest Format definition (3 places in the spec, 1 in `docs/examples/multi-protocol-agent.md`)
- **MCP mapping example**: `urn:air:example.com:mcp:finance-server` -> `urn:air:acme-corp.com:mcp:finance-server` (identifier, Trust Manifest identity, and the conceptual-mapping table row), so the URN publisher segment matches the entry's actual publisher and URLs
- **Well-Known URI prose**: no longer implies Trust Manifests are retrieved from artifact URLs; they are carried inline on entries
- **`docs/getting-started.md`**: Level 2 requires the `host` object per the spec; serving at the well-known URI is optional. Reworded the sentence that claimed well-known serving alone reaches Level 2
- **`docs/guides/adding-trust.md`**: removed `"identityType": "did"` from two examples whose identity is a `urn:air:` URN (not a DID), and reworded "Three `relation` types are defined" to reflect that `relation` is an open string with common values

## Verification

- Spec builds cleanly with `tools/build_spec.py`
- `specification/respec-config.json` parses as valid JSON


tadasant (approved):
Thank you!

PR #78, opened 2026-07-14
fix: updated the missing notes in ADR-0014 for markdown agent skills

### LINKED ISSUE #71: application/agent-skills+md contradicts the registered-suffix claim and ADR 0014's own rationale
The known-types list in "Catalog Entry > `type`" includes:

> `application/agent-skills+md` - an Agent Skill defined in a standard Markdown file (the suffix `+md` is to be registered)

Two contradictions:

1. The very next paragraph claims "These values are designed to align with official IANA media type registration standards. Standard ecosystem types use registered structured syntax suffixes (`+json`, `+zip`, `+gzip`)." The `+md` suffix is not a registered structured syntax suffix, so the list contradicts its own registrability claim.

2. ADR 0014's rationale already resolved this differently: "Standard Markdown is represented by the generic, globally registered `text/markdown` media type. Because using `text/markdown` in isolation does not indicate the document is an Agent Skill ... we use the standard IANA `profile` parameter (`profile="urn:air:agent-skill"`) to cleanly declare the semantic profile of the document while remaining 100% standard-compliant." That decision never made it into the known-types list - the list (in both the ADR's own Decision section and the spec) still carries `application/agent-skills+md`.

The docs type table (`docs/guides/creating-a-catalog.md`, "Choosing a type") repeats `application/agent-skills+md` and would need the same update.

Suggested resolution: replace `application/agent-skills+md` with `text/markdown;profile="urn:air:agent-skill"` per the ADR 0014 rationale, or explicitly decide to pursue `+md` registration and fix the surrounding registered-suffix claim. Either way the ADR's Decision and Rationale sections should stop disagreeing with each other.

liady:
Fixes #71.

ADR 0014's rationale already settled on `text/markdown;profile="urn:air:agent-skill"` for markdown skills, but the known-types list still carried `application/agent-skills+md` - both in the spec and in the ADR's own decision section - contradicting the registered-suffix claim right below it (`+md` is not a registered structured syntax suffix).

This updates the three places that still list the old type: the spec's known-types list, the ADR 0014 decision list, and the type table in the creating-a-catalog guide.


darrelmiller:
I had an action item to go propose the registration of a +md structured suffix.  I will do that this week.

mindpower:
Yes, we decided to go with the the `application/agent-skills+md`

We need to clean up the text (spec+adr) regarding text/markdown;profile="urn:air:agent-skill" part.

darrelmiller:
Email sent for feedback on registering +md as  structured suffix https://mailarchive.ietf.org/arch/msg/media-types/uzz7rt9wXPWSx5r66Ba2CKPkNsE/

darrelmiller (approved):

mindpower:
I'll make the fix of the wording and merge in.

PR #79, opened 2026-07-14
fix: terminology alignment and ADR 0007/0010 follow-ups

### LINKED ISSUE #72: Terminology and wording alignment: 'media type' leftovers, discovery strictness, ADR 0007/0010 follow-ups
A batch of wording/terminology alignment items. None is individually large, but all four leave the spec disagreeing with itself or with a recorded ADR.

## 1. "Media type" leftovers from before ADR 0014

ADR 0014's stated motivation for renaming `mediaType` to `type` was that "using the name mediaType implies support for any valid IANA media type ... which is not the intent". But the Terminology section still defines a Catalog Entry as "identified by a media type", and Design Goal 2 still says each entry "MUST declare its artifact type using a media type". The `type` field definition and the MCP appendix note already use the correct "open type identifier" framing; the intro material should match.

## 2. Discovery step 4 is stricter than Location Independence allows

"Link Relation Discovery" step 4 treats a response as a catalog only "if the response has a media type of `application/ai-catalog+json`", but "Location Independence" makes serving that media type a SHOULD, and the docs explicitly accept `application/json` as fine in practice. A catalog served as `application/json` would pass the well-known flow but fail the link-relation flow. Step 4 should tolerate generic JSON media types (e.g. check for a JSON media type plus a valid `specVersion`).

## 3. ADR 0010 follow-up: "AI Card" terminology entry is missing

ADR 0010's consequences say: "The spec should define 'AI Card' in its terminology section as the umbrella concept for any AI artifact metadata document." The Terminology section has no such entry.

## 4. ADR 0007 follow-up: url vs uri naming is inconsistent

ADR 0007's consequences say "The field is named `url` across all objects (catalog entries, collection references, attestations)", but the Attestation object uses `uri`, and the spec mixes suffix conventions overall: `documentationUrl`, `logoUrl`, `privacyPolicyUrl`, `termsOfServiceUrl` vs `governanceUri`, `registryUri`, `statementUri`, and attestation `uri`. Either normalize the field names, or update ADR 0007 to record the deliberate split (e.g. `url` for fetchable HTTP resources, `uri` where non-fetchable identifiers are allowed) so implementers know it is intentional.

liady:
Fixes #72.

Four small alignment items, none of them changing behavior on their own:

- The intro, the Catalog Entry terminology definition, and Design Goal 2 still described entries as identified by a "media type". ADR 0014 renamed the field to `type` precisely because it is an open type identifier, and the field definition already uses that framing, so the intro material now matches.
- Link relation discovery step 4 only accepted `application/ai-catalog+json` responses, while Location Independence makes serving that media type a SHOULD. A catalog served as `application/json` would pass the well-known flow but fail the link-relation flow. Step 4 now accepts generic JSON media types as long as the document has a valid `specVersion`.
- Added the "AI Card" terminology entry that ADR 0010 called for.
- ADR 0007 said the field is named `url` across all objects, but the spec uses `uri` where non-fetchable identifiers are allowed (attestation `uri`, `governanceUri`, `registryUri`, `statementUri`). Added an update note to the ADR recording that the split is deliberate, so implementers don't read it as drift.

### DIFF
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -10,7 +10,7 @@ and reducing interoperability.
 
 This document defines the **AI Catalog**: a typed, nestable JSON
 container for discovering heterogeneous AI artifacts. Each entry
-declares its artifact type via a type identifier and may reference or
+declares its artifact type via a type identifier (which is an IANA media type) and may reference or
 embed the native artifact metadata. A minimal catalog is simply a
 list of entries — names, types, and URLs — requiring no additional
 infrastructure.
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -56,13 +56,6 @@ Artifact
   manifest, an A2A agent card, a Claude Code plugin, a
   dataset descriptor, or a nested AI Catalog.
 
-AI Card
-: The umbrella concept for any AI artifact metadata document, such as
-  an A2A Agent Card, an MCP Server Card, or an Agent Skill. "AI Card"
-  is the name of the working group and the overall effort; the
-  normative term for such a document in this specification is
-  Catalog Entry.
-
 # Design Goals
 
 1. **Artifact Agnosticism**: The catalog MUST be capable of indexing
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -10,7 +10,7 @@ and reducing interoperability.
 
 This document defines the **AI Catalog**: a typed, nestable JSON
 container for discovering heterogeneous AI artifacts. Each entry
-declares its artifact type via a media type and may reference or
+declares its artifact type via a type identifier and may reference or
 embed the native artifact metadata. A minimal catalog is simply a
 list of entries — names, types, and URLs — requiring no additional
 infrastructure.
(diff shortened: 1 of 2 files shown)


specification/ai-catalog.md
```diff
@@ -10,7 +10,7 @@ and reducing interoperability.
 
 This document defines the **AI Catalog**: a typed, nestable JSON
 container for discovering heterogeneous AI artifacts. Each entry
-declares its artifact type via a media type and may reference or
+declares its artifact type via a type identifier and may reference or
```
darrelmiller, line 13: The value of `type` is still an IANA media type.  We are sticking with this value domain because then we don't need to define our own syntax and range of valid values.  There is precedence for using `type` as an indicator of media type in the IETF Web Linking specification https://datatracker.ietf.org/doc/html/rfc8288#section-3.4.1

specification/ai-catalog.md
```diff
@@ -56,14 +56,21 @@ Artifact
   manifest, an A2A agent card, a Claude Code plugin, a
   dataset descriptor, or a nested AI Catalog.
 
+AI Card
```
darrelmiller, line 59: I'm not sure we really want to keep the terminology of AI Card.  The repo was renamed and the org hasn't been changed primarily because we didn't want to break links.  We have been using "catalog entry" as the primary term.

specification/ai-catalog.md
mindpower: +1. We can word like this: 
   via a type identifier (which is IANA registered media type)

specification/ai-catalog.md
mindpower: Yes let's remove the AI card change part.

specification/ai-catalog.md
darrelmiller:
```suggestion
declares its artifact type via a type identifier (which is an IANA media type) and may reference or
```

darrelmiller:
From the Aug 20th meeting, there is consensus around this PR with the exception of the section about AI Card which we should remove.

Tehsmash (approved):

PR #77, opened 2026-07-10
Resolve #58: Add ADR-0017 and migrate to namespace extensions map

mindpower:
### Summary
This PR formally resolves Issue #58 (Proposal for JSON-LD `@context` Namespaces) by implementing the consensus reached during the July 2 and July 9 Working Group meetings and also comments on that issue.
 
To preserve the AI-Catalog's minimalist parser and avoid the complexity of JSON-LD graph resolution, the TSC agreed to reject the `@context` proposal. Instead, we are adopting a **loosely coupled architecture** with the Agentic Resource Discovery (ARD) specification and replacing the legacy `metadata` property with a strict `extensions` array.

### Changes Included
1. **Added `ADR-0017`**: Documents the decision to loosely couple ARD and AI-Catalog, preserving AI-Catalog as a strictly-typed "publishing format" while allowing ARD to evolve independently as a rich "discovery format."
2. **Updated `ai-catalog.md`**: 
   - Removed the legacy `metadata` property from the Catalog Entry, Trust Manifest, and root schemas.
   - Introduced the `extensions` array mechanism.
   - Rewrote the Extensibility section to enforce reverse-DNS/URL keys for extension boundaries.

### Motivation
This architecture allows AI-Catalog to remain a lightweight, fast "thin pointer," while freeing downstream aggregators (like ARD) to build the complex semantic envelopes they need for federated discovery without forcing those requirements into the base protocol.

Closes #58

PR #81, opened 2026-07-14
fix: make the Trust Manifest MUST rules satisfiable by the spec's own examples

### LINKED ISSUE #70: Trust Manifest MUST rules cannot be satisfied by open identifier formats or the spec's own appendix examples
Two normative MUSTs in the Trust Manifest section cannot be satisfied by the spec's own appendix examples, because both rules assume structure that the open formats do not guarantee.

## 1. Identity binding is undefined for non-URN identifiers

"Trust Manifest > Identity" says:

> the `identity` field's authority or trust domain MUST align with the publisher domain segment of the entry's URN `identifier` field. ... Consumers MUST reject a Trust Manifest whose identity domain does not align with the publisher domain of the entry's `identifier`.

But `identifier` is explicitly an open text format (any URI or URN; the `urn:air` form is only recommended, and custom formats are expected for closed or local systems). The "Mapping to Claude Code Plugins Marketplace" appendix itself uses `urn:claude-plugin:adspirer:ads-agent` style identifiers, which have no domain segment at all - so a conformant consumer following the MUST would have to reject the Trust Manifests in the spec's own examples.

The binding rule needs a scoping clause: e.g. it applies when the identifier uses the `urn:air` form (or another domain-anchored scheme), and consumers fall back to other verification (publisher attestations, signatures) otherwise.

## 2. Digest rejection rule vs the appendix's sha1 examples

"Verification Procedures > Digest Format" says: "Consumers MUST reject digest values using algorithms shorter than SHA-256."

The same Claude Plugins appendix maps the marketplace's `source.sha` (a git commit SHA) to `trustManifest.provenance[].sourceDigest` as `"sha1:..."`. Two problems:

- The examples violate the MUST: consumers are required to reject exactly the digests the appendix shows.
- A git commit SHA is a revision identifier, not a content digest of anything a consumer can fetch and hash, so mapping it to `sourceDigest` is arguably a category error regardless of algorithm (related to the artifact-integrity issue).

Suggested resolution: either carve out revision identifiers from the digest rule (e.g. a separate `sourceRevision` field), or change the appendix mapping so `sha` lands in `metadata` rather than `sourceDigest`.

liady:
Fixes #70.

Two normative MUSTs could not be satisfied by the Claude Plugins appendix examples:

1. The identity binding rule assumed every `identifier` has a publisher domain segment, but `identifier` is an open format and the appendix uses `urn:claude-plugin:...` identifiers with no domain at all. A consumer following the rule as written would have to reject the Trust Manifests in the spec's own examples. The rule is now scoped to identifiers that carry a publisher domain segment (such as the `urn:air` form), with a fallback to other verification (attestations, signatures) for formats that don't.

2. Digest Format requires consumers to reject digests weaker than SHA-256, yet the appendix mapped the marketplace's git commit `sha` to `sourceDigest` as `sha1:...`. A commit SHA is a revision identifier, not a content digest of anything a consumer can fetch and hash, so the mapping now lands it in entry `metadata.sourceSha` instead, and the examples no longer carry `sourceDigest` values that consumers are required to reject.

A dedicated `sourceRevision` field on the Provenance Link (as floated in the issue) would be a schema addition and would need its own ADR, so it is intentionally out of scope here.

### DIFF
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -513,15 +513,16 @@ A Trust Manifest MUST contain:
   DID, SPIFFE ID, or URL; these are illustrative and the set of
   identity schemes is open.
 
-When a Trust Manifest appears within a Catalog Entry, the `identity`
-field's trust domain MUST align with the publisher domain in the
-containing entry's `identifier` field. This binding ensures
-trust claims are associated with the authorized publisher namespace even
-when `identity` and `identifier` use different URI schemes.
-Consumers MUST reject a Trust Manifest whose `identity` domain does not
-align with the publisher domain in the containing entry's `identifier`.
-The `identity` is carried here so domain binding is part of the signed
-payload, rather than inferred only from unsigned entry context.
+When a Trust Manifest appears within a Catalog Entry whose
+`identifier` uses the `urn:air` naming format, consumers MUST
+extract the publisher domain from the identifier and verify that
+the `identity` field's authority or trust domain aligns with it.
+Consumers MUST reject a Trust Manifest whose identity domain does
+not align with the extracted publisher domain. For all other
+identifier formats, the identifier is opaque to the trust
+verification process; consumers MUST NOT attempt to derive a
+publisher domain from it and SHOULD rely on other verification
+instead, such as attestations or signature verification.
 
 When a Trust Manifest appears on a Host Info object, `identity`
 SHOULD match the host's `identifier` field when present.


specification/ai-catalog.md
```diff
@@ -406,12 +406,17 @@ A Trust Manifest MUST contain:
   DID, SPIFFE ID, or URL; these are illustrative and the set of
   identity schemes is open.
 
-When a Trust Manifest appears within a Catalog Entry, the `identity`
-field's authority or trust domain MUST align with the publisher domain segment of
-the entry's URN `identifier` field. This binding ensures trust claims are
-cryptographically bound to the authorized publisher of the catalog entry.
-Consumers MUST reject a Trust Manifest whose identity domain does not
-align with the publisher domain of the entry's `identifier`.
+When a Trust Manifest appears within a Catalog Entry whose
```
darrelmiller, line 516: We should be more precise in saying that only when the urn:air format can the trust manifest extract the domain from the identifier.

olijboyd:
Reviewed this against the current head while bringing our verifier back into line with it. Two notes from the implementation side.

**The `sourceSha` relocation lands on `metadata`, which the schema no longer defines.**

ADR-0017 (#77, merged 30 July) moved custom data into the `extensions` map. The CDDL now carries `? extensions: { * text => any }` on AICatalog, CatalogEntry and TrustManifest, and there is no `metadata` member in the schema at all. The only surviving field references in the normative document are four in the Claude Code Plugins Marketplace appendix: `Catalog metadata.description`, `Entry metadata.homepage`, and the `metadata` object in both worked examples. This PR adds `metadata.sourceSha` to the mapping table and to both of those examples, so it grows the count rather than clearing it.

Given that #70's whole point is that the appendix examples must be satisfiable by the normative text, the relocation probably wants to be:

```json
"extensions": {
  "https://ai-catalog.org/extensions/metadata": {
    "homepage": "https://www.adspirer.com",
    "sourceSha": "aa70dbdbbbb843e94a794c10c2b13f5dd66b5e40"
  }
}
```

which also gives the existing `homepage` and `description` rows somewhere conformant to point at. The wider version of this is not yours to fix in this PR, so I have raised it separately in #103.

**The identity scoping clause still needs a stated extraction rule.**

Agreed with @rrrodzilla that the scoped MUST leaves a consumer without a defined test. Two things become undefined once "carries a publisher domain segment" is the trigger: how to extract that segment from `identifier`, and how to derive the trust domain from the `identity` URI. Two implementations must answer both identically or the MUST is not interoperable.

This is what we settled on in our verifier, offered as a starting point rather than a proposal:

- Publisher segment: only `urn:air:<publisher>:...` is treated as carrying one. Everything else takes the "does not apply" branch. That makes the trigger a positive test against a named format rather than a guess at whether a string looks domain-shaped, which matters because the guess is the thing that would differ between implementations.
- Identity domain: `did:web` yields the first colon-separated element with `%3A` decoded and the port retained, since the port is part of the origin a resolver contacts. `https` and `spiffe` yield host plus port. Every other scheme, `dns:` included, is rejected rather than parsed, because its authority is ambiguous and a permissive reading would name an authority the URI does not control.

The port is the detail that caught us. A port-bearing `did:web` must not align with the bare apex domain, and it does if you normalise the port away.

PR #91, opened 2026-07-24
feat(spec): add ADR-0021 to support multiple identities via `alsoKnownAs`

### LINKED ISSUE #52: [Proposal]: Support multiple identities per catalog entry
## Summary

Today a catalog entry can declare only **one** cryptographic identity through `trustManifest.identity` 
field. Real artifacts frequently have **more than one** verifiable identity at the same time, for example:

- a SPIFFE ID (`spiffe://acme.com/ns/finance/sa/finance-a2a-pod`) for runtime / workload identity, and
- a DID (`did:web:acme-corp.com`) for organizational / publisher-anchored identity.

There is no first-class place to list multiple identities or a viable approach to support it. Authors are forced
to put one identity in `identity` and smuggle the others through `attestations[]` or `metadata` fields, 
where consumers do not reliably look for them and cannot treat them as verifiable subject identities.

**TLDR;** support natively a way to declare multiple identities on an entry so a consumer can
**pick the identity scheme it prefers or is able to verify**, instead of being limited to a single
scheme chosen by the publisher.

## Use cases

1. **Multiple identities** - One entry can hold multiple identities, e.g. SPIFFE for runtime, DID for
   publisher, or ANS for naming.
2. **Multiple verification methods** - An artifact is reachable as both `did:web:acme-corp.com`
   and a `did:<name>`, and different relying parties resolve different methods.
3. **Migration / Rotation** - Publishers transitioning from one identity scheme to another need to 
   advertise both without breaking either set of consumers.
4. **Consumer preference / capability** - A relying party can only verify the schemes it supports;
   it needs to choose a verifiable identity rather than reject an entry whose single `identity`
   uses a scheme it cannot resolve.

## Impact

- **Interoperability:** Without a standard shape, every publisher invents its own attestation/metadata
  convention for "secondary" identities, and consumers cannot rely on any of them.

## Problem statement

The Trust Manifest models identity as a **single** value and a **single** type hint:

> `identity`: A string containing a globally unique URI that serves as the **primary subject identifier** for this artifact.
> `identityType`: A string providing a type hint for the identity URI (e.g., "did", "spiffe", "dns", "ans").

Because there is exactly one `identity`, an artifact that legitimately holds two or more co-equal
identities cannot represent them.  The `Publisher` object has the same limitation.

The specification's own examples already show the workaround via dual-protocol finance agent
which declares its identity via:

```json
"trustManifest": {
  "identity": "spiffe://acme.com/ns/finance/sa/finance-a2a-pod",
  "identityType": "spiffe",
  "attestations": [
    {
      "type": "publisher-identity",
      "uri": "https://trust.acme.com/certs/publisher.jwt",
      "description": "Verifies did:web:acme-corp.com as publisher"
    }
  ]
}
```

Here the artifact really has **two** identities (a SPIFFE identity and a publisher-anchored
`did:web:acme-corp.com` identity), but only the SPIFFE ID is a first-class identity.
The DID only surfaces indirectly through `publisher.identifier` and `attestation` fields.

This is exactly the hacky usage of trust attestations to embed identity details that
motivates this issue.

### Why the workarounds are inadequate

- **Attestations are limiting.** An `Attestation` is a typed *claim with evidence*
  designed for compliance documents and proof artifacts, not for declaring an
  additional **identity** the consumer can pin and verify. There is no standard
  attestation `type` that means "this is another identity of the same subject",
  so consumers cannot programmatically discover alternate identities or know
  which URI to verify against.
- **`metadata` is opaque and non-interoperable.** As per spec, consumers
  "SHOULD ignore" metadata keys they do not recognize, so a second identity
  placed in `metadata` is invisible to clients and defeats the purpose.
- **Conflicts with the domain-alignment rule.** The spec requires the `identity`
  trust domain to align with the publisher domain of the entry's `identifier`.
  With two identities in different trust domains (e.g. `spiffe://acme.com/...`
  and `did:web:acme-corp.com`) or of different types, the rule is written
  for a single identity and gives no guidance on how multiple identities
  should each be bound.

## Proposed solutions

Below are only some examples of supporting multiple identities. This issue is not about finding the
right approach, but bringing awareness around the problem with potential approaches for resolution.

### `alias`/`alsoKnownAs`: a list of equivalent identity URIs

```json
"trustManifest": {
  "identity": "did:web:acme-corp.com:agent:finance",
  "identityType": "did",
  "alsoKnownAs": [
    "spiffe://acme.com/ns/finance/sa/finance-a2a-pod"
  ]
}
```

### `identities[]`: an array of Identity objects

```json
"trustManifest": {
  "identity": "did:web:acme-corp.com:agent:finance",
  "identityType": "did",
  "identities": [
    {
      "identity": "spiffe://acme.com/ns/finance/sa/finance-a2a-pod",
      "identityType": "spiffe",
      "signature": "eyJhbGciOiJFUzI1NiJ9..spiffe-signed"
    },
    {
      "identity": "did:key:z6Mk...",
      "identityType": "did",
      "signature": "eyJhbGciOiJFUzI1NiJ9..didkey-signed"
    }
  ],
  "signature": "eyJhbGciOiJFUzI1NiJ9..primary-signed"
}
```

### `trustManifests[]`: multiple Trust Manifests per entry

For identities that live in genuinely different trust domains and need fully independent trust metadata
(separate attestations, provenance, signatures), allow an array of Trust Manifests on an entry.

ramizpolic:
## Summary

This PR resolves #52 (support for multiple identities per artifact) by adding an
OPTIONAL `alsoKnownAs` member to the Trust Manifest, following the approach 
favored in the issue discussion and AI Catalog meeting over an `identities[]` array.

Real artifacts frequently hold more than one verifiable identity at the same
time, for example a SPIFFE ID for runtime workload identity and a DID for
publisher-anchored organizational identity. Today there is no first-class
place to declare them, forcing publishers to embed secondary identities
through `attestations[]` or `metadata`, where consumers cannot discover or
verify them.

## Changes Included

1. Added `ADR-0019`: Documents the decision to declare alternative identities
   through an `alsoKnownAs` URI array inside the Trust Manifest (rather than
   on the Catalog Entry), so aliases are covered by the existing manifest
   signature and remain verifiable through the trust bundle.
2. Updated `ai-catalog.md`:
   - Added the `alsoKnownAs` member to the Trust Manifest optional members,
     with normative rules (canonical `identity`, no duplicates,
     domain-alignment applies only to `identity`).
   - Added a "Verifying Alternative Identities" verification procedure:
     verifying the manifest `signature` verifies every listed alias, no
     per-alias proof mechanism is required.
   - Updated the CDDL schema, data model diagram, and examples.

## Motivation

Consumers can programmatically discover and pin alternate identities of the
same subject using whichever identity scheme their tooling resolves, while
`identity` remains the single canonical identifier for referencing and
equivalence checking. Because `alsoKnownAs` lives inside the signed manifest
content, no new signing mechanism is needed and the field is tamper-proof
under the catalog-poisoning threat model.

### DIFF
--- a/adr/0021-multiple-identities-alsoknownas.md
+++ b/adr/0021-multiple-identities-alsoknownas.md
@@ -1,4 +1,4 @@
-# ADR 0019: Support Multiple Identities via `alsoKnownAs` in the Trust Manifest
+# ADR 0021: Support Multiple Identities via `alsoKnownAs` in the Trust Manifest
 
 ## Status
 Proposed
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -570,9 +570,9 @@ present.
 The following members are OPTIONAL:
 
 `identityType`
-: A string providing a type hint for the identity URI (e.g., "did",
+: A string providing a type hint for the `identity` URI (e.g., "did",
   "spiffe", "dns"). This field is OPTIONAL when the type is evident
-  from the URI scheme.
+  from the URI scheme. It does not describe entries of `alsoKnownAs`.
 
 `alsoKnownAs`
 : An array of strings, each containing a globally unique URI
@@ -587,6 +587,9 @@ The following members are OPTIONAL:
     - The domain-alignment rule defined in [Identity](#identity)
       applies only to `identity`. Aliases MAY belong to different
       trust domains or identity schemes.
+    - The `identityType` hint applies only to `identity`; the type of
+      an alias is inferred from its URI scheme (e.g., `did:`,
+      `spiffe:`).
     - Aliases are publisher claims covered by the Trust Manifest
       `signature`; no per-alias proof is required. Consumers MUST NOT
       rely on an alias from a manifest whose signature is absent or
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -528,12 +528,13 @@ SHOULD match the host's `identifier` field when present.
 
 An artifact may legitimately hold more than one verifiable identity at
 the same time, for example a SPIFFE ID for runtime workload identity
-and a DID for publisher-anchored organizational identity. Additional
-identities are declared through the OPTIONAL `alsoKnownAs` member (see
-[Optional Members](#optional-members)). The `identity` field remains the
-single canonical subject identifier: consumers MUST use `identity` when
-referencing the artifact's trust subject or checking identity
-equivalence between entries.
+and a DID for publisher-anchored organizational identity, or it may
+retain a former identifier after a rename or migration between
+identities or schemes. Additional identities are declared through the
+OPTIONAL `alsoKnownAs` member (see [Optional Members](#optional-members)). 
+The `identity` field remains the single canonical subject identifier:
+consumers MUST use `identity` when referencing the artifact's trust
+subject or checking identity equivalence between entries.
 
 When multiple entries share the same `identifier` (with different `version`
 values), each entry MAY carry its own Trust Manifest. There is no
@@ -572,24 +573,38 @@ The following members are OPTIONAL:
 `identityType`
 : A string providing a type hint for the `identity` URI (e.g., "did",
   "spiffe", "dns"). This field is OPTIONAL when the type is evident
-  from the URI scheme. It does not describe entries of `alsoKnownAs`.
+  from the URI scheme.
 
 `alsoKnownAs`
-: An array of strings, each containing a globally unique URI
-  [[RFC3986]] that identifies the same subject as `identity` under an
-  alternative identity scheme. The following rules apply:
+: An array of Identity Alias objects, each asserting an identity of
+  the same subject as `identity` under an alternative identity
+  scheme. An Identity Alias object MUST contain:
 
-    - An alias is an equivalent identity of the subject; `identity`
-      alone remains canonical and is used for referencing and
-      equivalence checking.
-    - The array is an unordered set: it MUST NOT contain duplicate
-      values or the value of the `identity` field.
+    `identity`
+    : A string containing a globally unique URI [[RFC3986]] that
+      identifies the same subject as the manifest's canonical
+      `identity`.
+
+    An Identity Alias object MAY contain:
+
+    `identityType`
+    : A string providing a type hint for the alias `identity` URI
+      (e.g., "did", "spiffe", "dns"). This member is OPTIONAL when
+      the type is evident from the URI scheme.
+
+    The following rules apply:
+
+    - An alias is an equivalent identity of the subject; the
+      manifest's `identity` alone remains canonical and is used for
+      referencing and equivalence checking. Aliases MAY also record
+      former identifiers of the subject, preserving continuity after
+      a rename or a migration between identities or schemes.
+    - The array is an unordered set: it MUST NOT contain two aliases
+      with the same `identity` value, and no alias `identity` may
+      equal the manifest's canonical `identity`.
     - The domain-alignment rule defined in [Identity](#identity)
-      applies only to `identity`. Aliases MAY belong to different
-      trust domains or identity schemes.
-    - The `identityType` hint applies only to `identity`; the type of
-      an alias is inferred from its URI scheme (e.g., `did:`,
-      `spiffe:`).
+      applies only to the canonical `identity`. Aliases MAY belong to
+      different trust domains or identity schemes.
     - Aliases are publisher claims covered by the Trust Manifest
       `signature`; no per-alias proof is required. Consumers MUST NOT
       rely on an alias from a manifest whose signature is absent or
@@ -653,7 +668,10 @@ provenance:
   "identity": "did:web:acme.com:agent:finance",
   "identityType": "did",
   "alsoKnownAs": [
-    "spiffe://acme.com/ns/finance/sa/finance-agent-pod"
+    {
+      "identity": "spiffe://acme.com/ns/finance/sa/finance-agent-pod",
+      "identityType": "spiffe"
+    }
   ],
   "trustSchema": {
     "identifier": "urn:trust:acme-enterprise-v1",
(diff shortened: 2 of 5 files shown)


kthota-g:
I feel there is inconsistency with type discovery for `identity` and `alsoKnownAs`. For `identity`, publishers can provide an explicit `identityType` hint. But for `alsoKnownAs`, the consumers are forced to rely on inferring the type from URI scheme. At least I felt there is a mismatch between `identityType` and `alsoKnownAs` looking at the manifest.

If we don't want to remove `identityType` or introduce `identityType` per alias, probably explicitly define that `identityType` is a type hint only for `identity` attribute.

ramizpolic:
thanks for raising this @kthota-g. i have patched the spec to include details around type inference for alias identities, hope it addresses your concerns.

darrelmiller:
@ramizpolic Can you clarify why we don't need identityType?  

Can we also clarify if the purpose is both to be an alternative identify, and a record of a rename. i.e. the "alsoKnownAs" includes "formerlyKnownAs".

specification/ai-catalog.md
```diff
@@ -683,6 +717,27 @@ To verify the publisher of an artifact:
...
+[Trust Manifest Signatures](#trust-manifest-signatures) also verifies
+every listed alias: a valid signature proves that the publisher
+controlling the canonical `identity` claims each alias as an
+equivalent identity of the subject. Consumers MUST NOT rely on aliases
+from a manifest whose signature is absent or fails verification.
+
+Once the signature is verified, a consumer MAY use whichever identity,
+canonical or alias, matches the identity schemes it is able to
+resolve — for example, using a DID alias for DID-based discovery when
+its tooling cannot resolve a canonical SPIFFE ID.
```
yangbaechu, line 734: The example's consumer can't resolve the canonical SPIFFE ID, but key resolution is based on `identity`. So it can't verify the signature the alias depends on. Would a case where `identity` resolves fit better?

jdamick:
This is something that would be valuable to my use case. We would like to have the flexibility to issue a DID and SPIFFE id. There different preferences depending on the use and who is using the ai-catalog that may require DID or SPIFFE and I've heard requests for each.

adr/0021-multiple-identities-alsoknownas.md
```diff
@@ -0,0 +1,78 @@
...
+A key constraint shaped this decision: **declared identities MUST be verifiable through the trust bundle.** An identity claim a consumer cannot verify is worse than no claim at all.
+
+## Decision
+The Trust Manifest gains an OPTIONAL `alsoKnownAs` member: an array of globally unique URIs, each asserting an alternative identity of the **same subject** as the canonical `identity` field.
+
+```json
+"trustManifest": {
+  "identity": "spiffe://acme.com/ns/finance/sa/finance-agent-pod",
+  "identityType": "spiffe",
+  "alsoKnownAs": [
```
jdamick, line 34: does this need a type or at least the possibility of a type?

Perhaps:
```
"alsoKnownAs" : [
  { 
   "identity": "did:web:acme-corp.com:agent:finance",
   "identityType": "did:web",
  }
]
```

specification/ai-catalog.md
ramizpolic: Good catch, thanks @yangbaechu! Added a fix to the spec to state explicitly that signature verification always resolves the key from the canonical identity (so a consumer that can't resolve it can't rely on any alias) and patched the examples.

adr/0021-multiple-identities-alsoknownas.md
ramizpolic: Thanks for raising this @jdamick. I patched the spec to use an array of Identity Alias objects mirroring the canonical identity/identityType pair. This also covers your use case as consumers can pick the alias by explicit type rather than inferring from schemes (which is not always possible as described in  https://github.com/Agent-Card/ai-catalog/pull/91#issuecomment-5440541288).

ramizpolic:
> @ramizpolic Can you clarify why we don't need identityType?
> 
> Can we also clarify if the purpose is both to be an alternative identify, and a record of a rename. i.e. the "alsoKnownAs" includes "formerlyKnownAs".

Good points @darrelmiller, addressed both of the questions in the spec:

- **identityType**: we now have it per alias since pure scheme inference was ambiguous for schemes shared by multiple identity types (e.g., an https: URI could be a JWK Set or a did:web equivalent)
- **renames**:  patched the spec to include notes about former identifiers, preserving continuity after a rename or migration between identities and schemes

Wolfe-Jam:
**Proposal:** Registry-scoped identity + `alsoKnownAs` together solve both determinism and flexibility.

**The Pattern:**

Make identity deterministic from registry. Let artifacts declare schemes within each registry:

```
Registry 1 (ai-catalog.io):
  Canonical: ai-catalog.io/anthropic/finance-agent
  alsoKnownAs: ["did:web:anthropic.com/finance-agent", "spiffe://anthropic.com/finance"]

Registry 2 (other-registry.io):
  Canonical: other-registry.io/anthropic/finance-agent
  alsoKnownAs: ["did:web:anthropic.com/finance-agent", "spiffe://anthropic.com/finance"]
```

Same artifact, different canonical identities per registry. Within each, the artifact declares which schemes it recognizes.

**Why Together:**

- `alsoKnownAs` alone: Identity floats free; consumers don't know which registry is canonical.
- Registry-scoped alone: Loses flexibility for multi-scheme artifacts.
- Both: Canonical identity is deterministic (registry-bound). Aliases are flexible (publisher-signed). Threat models align — registry proves canonical, manifest proves schemes.

**Media-Types are definitive, yet No Media-Types Required:**

You don't need IANA registration to implement registry-scoped identity. You need:
1. One registry of truth per artifact class
2. Canonical identity scoped by registry + namespace
3. Aliases declared within, signed by publisher (your ADR-0021)

Lower bar, same problem solved.

jdamick:
> **Proposal:** Registry-scoped identity + `alsoKnownAs` together solve both determinism and flexibility.
> 
> **The Pattern:**
> 
> Make identity deterministic from registry. Let artifacts declare schemes within each registry:
> 

I'm not sure I understand this proposal because the entries or the trustManifest have the identifier, not the registry.  Why would you promote the identifier to the registry level?  I don't see how that addresses what is being discussed here?

Wolfe-Jam:
You're right. The trustManifest has the identifier; the signature covers it. alsoKnownAs with manifest signature already solves the multi-scheme problem. I was proposing a registry-level constraint where you already have one.

kthota-g:
An artifact's identity has no bearing on the catalog that's hosted in. For example: A finance-agent published by acme.com and its `trustManifest.identity` would remain the same across any catalogs/registries and consumers can verify just the artifact's identity based on the trustManifest.

Although catalog host's identity itself can be verified via `host.identifer`.

jdamick (approved):

muscariello:
This PR claims to resolve #52  ("support multiple identities per catalog entry"), but alsoKnownAs only solves the narrower case:

one principal, multiple labels for itself. #52's own discussion already considered the multi-principal case (independent signers, via an identities[]-style array) and dropped it on @darrelmiller's objection: "consumers would have to choose which identity to reference the entry by." 

#108  (ahead of this one in the queue, approved pending a version-symmetry fix) resolves that objection independently: entry.identifier is now explicitly the reference, no signer's identity competes for it.

That matters because the flagship example here (SPIFFE workload + DID publisher) is itself a multi-principal case, not a same-principal alias — different keys, different trust roots, and the canonical-key-only verification model can't actually deliver "verify via whichever scheme you support" for it. So the PR's stated scope (#52 in full) isn't actually covered yet.

Given #108 removes the original blocker, should this PR's scope be extended to add the independently-signed multi-principal case it was meant to resolve? 

I think it should more so if the merge queue is approved as it seems it is already, and I agree with that.
#109 #108 #110 #105 

Fix #91 based on the merged PRs above and solve #52 entirely.

@jonathanhefner @mindpower @ramizpolic @darrelmiller @kthota-g

PR #96, opened 2026-08-06
feat(spec): add Deployment official extension (resolves #64)

### LINKED ISSUE #64: Spec Proposal: Deployment metadata
### NOTE:  
This issue was originally posted on the [ards-project ](https://github.com/ards-project/ard-spec/) repo, issue [#44](https://github.com/ards-project/ard-spec/issues/44). Moving here for visibility. Please refer to the original issue for already existing community feedback and comments

The issue was also referred to in other issues:

https://github.com/ards-project/ard-spec/issues/53
https://github.com/ards-project/ard-spec/issues/55
https://github.com/ards-project/ard-spec/issues/44#event-8070795329
https://github.com/ards-project/ard-spec/issues/57

**The Problem:**

AI catalog conflates "what the agent is" (identity) with "where it lives" (instance). This is fine in the open internet but doesn't scale that well for assets living inside the enterprise:

- The SAME logical agent runs in dev, staging, and prod
- The SAME logical agent has replicas in us-east, eu-west, ap-southeast (this may be hidden behind a load balancer, but not always)
- These are NOT different agents - they're different instances of the same agent

The geolocation comes with additional concerns like compliance requirements and data residency restrictions. The point could be made 
that this is no different from auth and should be delegated to each assets card or modeled through metadata. However:

- None of the standard cards have these concepts
- Metadata without standardization is suboptimal
- It hinders the agent's ability to safely install new capabilities at runtime

**Proposal: Structured deployment metadata**

- Keep the `url` attribute as the default entry point for the resource
- Enrich the entry with an optional `instances` array which lists all valid URLs alongside deployment information such as:
  - environment
  - region
  - data residency
  - compliance
  
```json
{
  "identifier": "urn:ai:acme.com:finance:invoice-processor",
  "displayName": "Invoice Processor",
  "url": "https://api.acme.com/agents/invoice",
  "type": "application/a2a-agent-card+json",
  "instances": [
    {
      "instanceId": "invoice-prod-us",
      "environment": "production",
      "url": "https://api.acme.com/agents/invoice",
      "region": "us-east-1"
    },
    {
      "instanceId": "invoice-prod-eu",
      "environment": "production",
      "url": "https://eu-api.acme.com/agents/invoice",
      "region": "eu-west-1",
      "dataResidency": ["EU"],
      "compliance": "GDPR"
    }
  ]
}
```
This makes it possible to create deterministic guardrails to make sure agents only select assets that conform to their own compliance rules. 

**Discovery Value:**

- Test discovery in staging without prod impact
- Discover geo-appropriate instances (EU clients → EU instances for data residency)
- Support canary discovery (discover beta instance for testing/shadowing)
- Maintain single logical identity across deployment pipeline

**Alternatives Considered:**

- **Simply use different catalogs:** If my agent can only use GDPR compliant assets, then just point it to a previously curated/filtered catalog
     - While a valid solution, it's less flexible: What if GDPR is only required for _some_ operations and only for clients associated to a specific geography? This approach overly complicates management and can lead to inconsistencies

marianogonzalez:
## Summary

This PR resolves [#64](https://github.com/Agent-Card/ai-catalog/issues/64)
(structured deployment metadata) by adding an **official extension** —
`https://ai-catalog.org/extensions/deployment` — rather than a native
`instances` field on the Catalog Entry as the original proposal suggested.

A Catalog Entry conflates *identity* ("what the artifact is") with *instance*
("where it lives"). The same logical agent or server is routinely deployed
many times — across environments (development, staging, production), across
release channels (stable, beta, LTS, edge), and across regions (`us-east-1`,
`eu-west-1`) whose data-residency and compliance constraints differ. The
entry's `identifier` and top-level `url` capture the artifact and its default
entry point but cannot express these deployments. The extension lets a
publisher enumerate them on an entry so a consumer can deterministically
select an instance that satisfies its policy — an EU client targeting a GDPR
instance in `eu-west-1`, a test harness targeting `staging`, or a client
opting into a `beta` release channel — while the entry keeps a single logical
identity.

Because the extension is additive and unrecognized `extensions` keys are
ignored, it needs **no `specVersion` change** — a catalog carrying it remains
a conformant `1.0` document.

## Changes Included

1. **`adr/0022-deployment-metadata-extension.md`** (new) — records the
   decision to implement #64 as an official extension rather than a native
   entry field, consistent with the closed-core philosophy of
   [ADR-0012](adr/0012-extensibility-via-metadata.md).
2. **`specification/ai-catalog.md`**:
   - Registered the extension in the **Official Extensions** list alongside
     `metadata`.
   - Added the **Deployment Metadata Extension** section: value structure
     (a REQUIRED, non-empty `instances[]` array), the Instance object
     (`instanceId` and `url` REQUIRED; `environment`, `releaseChannel`,
     `region`, `dataResidency`, `compliance`, `description` OPTIONAL),
     selection semantics with a no-unsafe-fallback rule, and an example.
   - Added an informative **CDDL** block for the extension value.
   - Added a **Privacy Considerations** note on infrastructure-topology
     disclosure and unsigned producer assertions.
3. **Developer docs** — a new annotated example
   (`docs/examples/deployment-metadata.md`), an authoring section in
   *Creating a Catalog*, an instance-selection section in *Consuming
   Catalogs*, plus cross-links and a nav entry.

## Reconciling the original proposal

The proposal in #64 predates several current conventions; the extension uses
today's forms:

| Proposal (outdated) | This PR (current) |
|---|---|
| `urn:ai:…` | `urn:air:{publisher}:{namespace}:{name}` (ADR-0015) |
| native `instances` field on the entry | official `extensions` entry (keeps the core schema closed) |
| `compliance` as a bare string (`"GDPR"`) | `compliance` as a string array, matching `dataResidency` and allowing multiple regimes |
| "canary" discovery | `environment` (dev/staging/production) kept distinct from a new `releaseChannel` track (stable/beta/LTS/edge), since canary is a release strategy, not an environment tier |

## Design notes

- **Identity vs. instance.** Every instance is an alternative endpoint for the
  *same* `identifier`; instances are deployments, not distinct artifacts. The
  entry's top-level `url` remains the default entry point, and one instance
  `url` SHOULD equal it so the default is represented among the instances.
- **`environment` and `releaseChannel` are orthogonal.** `environment`
  describes *where* a version runs; `releaseChannel` describes the maturity
  track of the deployed version. Keeping them separate avoids overloading
  `environment` with release-strategy values.
- **Deployment metadata is not a trust control.** `dataResidency` and
  `compliance` are unsigned producer assertions unless carried in a signed
  Trust Manifest; a consumer that needs assurance MUST corroborate them via
  attestations, not via this extension.
- **No unsafe fallback.** A consumer that understands the extension MAY select
  an instance matching its policy; if none matches it SHOULD treat the
  artifact as unavailable for that request rather than fall back to a
  non-conforming instance — an EU-only client never silently uses a non-EU
  deployment.

## Interoperability impact

No change to interoperability expectations for existing catalogs. The core
schema, CDDL for core types, conformance levels, and `specVersion` are all
unchanged; a consumer that does not implement the extension ignores it without
error, exactly as for any unrecognized `extensions` key.

## Non-goals

- A deployment **resolver / load balancer** or health-checking behavior — this
  is discovery metadata only; the entry's `url` and each instance `url` remain
  ordinary references.
- Promoting deployment metadata to a **core field** — deliberately kept an
  extension per ADR-0012 / ADR-0022.
- Treating `compliance` / `dataResidency` as **enforceable guarantees** —
  they are producer assertions, verified (when needed) through the Trust
  Manifest.

## Validation

- `uv run --locked python tools/build_spec.py specification/ai-catalog.md dist/index.html --config specification/respec-config.json` — passes.
- `uv run --python 3.12 --locked --group docs mkdocs build --strict` — passes, no broken links.
- All deployment-bearing JSON examples validated.

### DIFF
--- a/docs/examples/deployment-metadata.md
+++ b/docs/examples/deployment-metadata.md
@@ -1,0 +1,145 @@
+# Example: Deployment Metadata
+
+A single logical agent often runs in more than one place: production and
+staging, `us-east-1` and `eu-west-1`, with different data-residency and
+compliance constraints in each region. AI Catalog keeps **identity** ("what
+the agent is") separate from **instance** ("where it lives"): the entry has one
+stable `identifier`, and the [Deployment official extension](../specification.md)
+enumerates the concrete deployment instances behind it.
+
+This lets a consumer deterministically pick an instance that satisfies its
+policy — an EU client selecting the GDPR instance in `eu-west-1`, a test harness
+targeting `staging`, or a client opting into a `beta` release channel — while
+everything stays under one logical identity.
+
+## The extension
+
+The extension key (namespace) is:
+
+```
+https://ai-catalog.org/extensions/deployment
+```
+
+It lives inside an entry's `extensions` map. Its value is an object with a
+required, non-empty `instances` array. Each instance describes one deployment of
+the **same** logical artifact — instances are deployments, not distinct
+artifacts.
+
+!!! note "The entry `url` stays the default"
+    The entry's top-level `url` remains the default entry point. Each instance's
+    `url` is an alternative endpoint for the same `identifier`. One instance
+    `url` should equal the entry `url` so the default is represented among the
+    instances.
+
+## Full example
+
+```json
+{
+  "identifier": "urn:air:acme-corp.com:agent:invoice-processor",
+  "type": "application/a2a-agent-card+json",
+  "url": "https://api.acme-corp.com/agents/invoice",
+  "tags": ["finance", "a2a"],
+  "extensions": {
+    "https://ai-catalog.org/extensions/deployment": {
+      "instances": [
+        {
+          "instanceId": "invoice-prod-us",
+          "environment": "production",
+          "url": "https://api.acme-corp.com/agents/invoice",
+          "region": "us-east-1"
+        },
+        {
+          "instanceId": "invoice-prod-eu",
+          "environment": "production",
+          "url": "https://eu-api.acme-corp.com/agents/invoice",
+          "region": "eu-west-1",
+          "dataResidency": ["EU"],
+          "compliance": ["GDPR"]
+        },
+        {
+          "instanceId": "invoice-staging",
+          "environment": "staging",
+          "url": "https://staging-api.acme-corp.com/agents/invoice",
+          "region": "us-east-1",
+          "releaseChannel": "beta"
+        }
+      ]
+    }
+  }
+}
+```
+
+## Instance fields
+
+Each object in the `instances` array supports these fields:
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `instanceId` | string | Required | Stable identifier for this instance, unique within the entry |
+| `url` | string | Required | Entry-point URL for this instance |
+| `environment` | string | Optional | Deployment environment, e.g. `production`, `staging`, `development` |
+| `releaseChannel` | string | Optional | Release maturity track, orthogonal to `environment`, e.g. `stable`, `beta`, `LTS`, `EDGE` |
+| `region` | string | Optional | Deployment region, e.g. `us-east-1`, `eu-west-1` |
+| `dataResidency` | string[] | Optional | Jurisdictions where data is stored or processed, e.g. `["EU"]`, `["US","CA"]` |
+| `compliance` | string[] | Optional | Compliance regimes the instance conforms to, e.g. `["GDPR"]`, `["SOC2","HIPAA"]` |
+| `description` | string | Optional | Human-readable label for the instance |
+
+## Selecting an instance
+
+A consumer that understands the extension may select an instance whose
+`environment`, `region`, `dataResidency`, or `compliance` satisfies its policy.
+If no instance matches, it **should not** fall back to a non-conforming instance.
+
+```python
+DEPLOYMENT_KEY = "https://ai-catalog.org/extensions/deployment"
+
+def select_instance(entry, environment=None, release_channel=None,
+                    region=None, compliance=None):
+    """Pick a deployment instance matching the consumer's policy.
+
+    Returns None when no instance conforms — don't fall back
+    to a non-conforming instance (e.g. the entry url) in that case.
+    """
+    ext = entry.get("extensions", {}).get(DEPLOYMENT_KEY)
+    if not ext:
+        # Extension unknown/absent: fall back to the entry's default url.
+        return {"url": entry["url"]}
+
+    for inst in ext["instances"]:
+        if environment and inst.get("environment") != environment:
+            continue
+        if release_channel and inst.get("releaseChannel") != release_channel:
+            continue
+        if region and inst.get("region") != region:
+            continue
+        if compliance and compliance not in inst.get("compliance", []):
+            continue
+        return inst
+
+    return None  # no conforming instance — do not fall back
+```
+
+```python
+entry = ...  # the invoice-processor entry above
+
+# EU client bound by GDPR: resolves to the eu-west-1 instance.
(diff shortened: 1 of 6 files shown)


docs/examples/deployment-metadata.md
```diff
@@ -0,0 +1,145 @@
...
+          "instanceId": "invoice-staging",
+          "environment": "staging",
+          "url": "https://staging-api.acme-corp.com/agents/invoice",
+          "region": "us-east-1",
+          "releaseChannel": "beta"
+        }
+      ]
+    }
+  }
+}
```
Tehsmash, line 69: I would consider using the ai-catalog spec itself to solve this problem for example:
```
{
  "identifier": "urn:air:acme-corp.com:agent:invoice-processor",
  "type": "application/ai-catalog+json",
  "tags": ["finance", "a2a"],
  "data": {
    "entries": [
      {
        "id": "urn:air:acme-corp.com:agent:invoice-processor-us",
        "type": "application/a2a-agent-card+json",
        "url": "https://api.acme-corp.com/agents/invoice",
        "extensions": {
          "https://ai-catalog.org/extensions/deployment": {
              "environment": "production",
              "region": "us-east-1"
          }
      },
      {
        "id": "urn:air:acme-corp.com:agent:invoice-processor-prod-eu",
        "type": "application/a2a-agent-card+json",
        "url": "https://eu-api.acme-corp.com/agents/invoice",
        "extensions": {
            "https://ai-catalog.org/extensions/deployment": {
                "region": "eu-west-1",
                "dataResidency": ["EU"],
                "compliance": ["GDPR"],
                "instanceId": "invoice-prod-eu",
                "environment": "production"
            }
        }
      },
      {
        "id": "urn:air:acme-corp.com:agent:invoice-processor-staging-eu",
        "type": "application/a2a-agent-card+json",
        "url": "https://staging-api.acme-corp.com/agents/invoice",
        "extensions": {
            "https://ai-catalog.org/extensions/deployment": {
                "instanceId": "invoice-staging",
                "environment": "staging",
                "region": "us-east-1",
                "releaseChannel": "beta"
            }
         }
      }
    ]
  }
}
```

I think on the concept of a default URL or something like that today in the ai-catalog-cli we have logic which is:

```
If ID resolves to type ai-catalog+json and the user hasn't requested a specific type:
   prompt the user to provide `--type` to give their preference 

if catalog contains more than one entry of --type:
	prompt the user with the list of available entries to pick from
```

So you gracefully degrade from:

```
ai-catalog-cli pull urn:air:acme-corp.com:agent:invoice-processor
```

to 

```
ai-catalog-cli pull --type application/a2a-agent-card+json  urn:air:acme-corp.com:agent:invoice-processor
```

to 

```
ai-catalog-cli pull urn:air:acme-corp.com:agent:invoice-processor-staging-eu
```

This removes any implict "default" and the client must pick explicitly.

If you want a "generic" or default then you could go for either adding a default entry to the catalog, or you could have a top level catalog which includes the "default" entries for different types and add a nested catalog for when you want to go to more specific endpoints.

The key point is that once a client has selected `urn:air:acme-corp.com:agent:invoice-processor-staging-eu` you can alway re-discover the same catalog entry.

marianogonzalez: Thank you for your feedback @Tehsmash . In my original proposal, this was indeed part of the entry rather than an extension. At the time, I received feedback that it should be instead be modeled as an extension in order to keep the specification simple.

More than happy to switch back if that's the consensus. We must be careful not to introduce any backwards compatibility issues in the process though

Tehsmash: What I described is still an extension, its just that it uses the catalog entries as the list, then added deployment metadata on each specific entry.

PR #99, opened 2026-08-12
fix: make Agent Plugin entries retrievable

jonathanhefner:
Catalog URLs are defined to return the complete artifact bytes, but manifest-only entries left consumers without a way to acquire plugin components.

Replace the JSON manifest type with ZIP and gzip bundle types.


darrelmiller (approved):

mindpower (approved):

tadasant (approved):

khou22 (commented):
LGTM thanks

PR #104, opened 2026-08-26
Replace legacy `metadata` guidance with `extensions`

### LINKED ISSUE #103: The guides still document `metadata`, which ADR-0017 replaced with `extensions`
ADR-0017 (#77, merged 30 July) replaced the `metadata` member with an `extensions` map, and the normative document followed: the CDDL carries `? extensions: { * text => any }` on AICatalog, CatalogEntry and TrustManifest, there is an Extensions section defining key naming and the official `https://ai-catalog.org/extensions/metadata` type, and `metadata` is no longer defined as a member anywhere in the schema.

The documentation did not follow. No file under `docs/` mentions `extensions` at all, and the published guides still teach `metadata` as the extension point:

- `docs/guides/creating-a-catalog.md` defines `metadata` in the field list, lists it in the entry field table, and carries a "Metadata extensibility" section whose example is `"metadata": { ... }`.
- `docs/guides/adding-trust.md` lists `metadata` in the Trust Manifest field table as the "open map for custom trust metadata".

Both are live: <https://ai-catalog.io/guides/creating-a-catalog/> and <https://ai-catalog.io/guides/adding-trust/>.

The consequence is that a publisher who follows the getting-started guide today produces a document the normative CDDL rejects, and does not discover it until something validates against the schema rather than against the guide.

Four further field references survive in the Claude Code Plugins Marketplace appendix of the specification itself (`Catalog metadata.description`, `Entry metadata.homepage`, and the `metadata` object in both worked examples). #81 currently adds `metadata.sourceSha` alongside them, which is where I noticed this.

Found while updating our Python implementation to head. Happy to open a PR moving the guides and the appendix onto `extensions`, using `https://ai-catalog.org/extensions/metadata` for the generic key-value cases, if that is wanted from outside the WG. Flagging rather than assuming, since it touches the getting-started surface.

jonathanhefner:
ADR-0017 replaced the native `metadata` member, but the authoring guides and Claude marketplace mapping still showed documents that conflict with the normative CDDL.

This PR documents `extensions` at the catalog, entry, and Trust Manifest scopes, and describes the official metadata extension (without imposing a second namespace convention on its schemaless payload).

This PR also omits selective Claude `description` and `homepage` mappings, and removes the obsolete Adspirer marketplace snapshot.

Closes #103.

### DIFF
--- a/docs/mappings/claude-code-plugins.md
+++ b/docs/mappings/claude-code-plugins.md
@@ -1,159 +1,183 @@
-# Mapping to Claude Code Plugins Marketplace
+# Mapping to Claude Code Plugin Marketplaces
 
-This guide describes how the Anthropic Claude Code Plugins
-marketplace format (see
-[claude-plugins-official](https://github.com/anthropics/claude-plugins-official))
-maps to AI Catalog, enabling Claude Code plugins to be discovered,
-indexed, and distributed through a unified catalog alongside other AI
-artifacts.
+This guide describes how a
+[Claude Code plugin marketplace](https://code.claude.com/docs/en/plugin-marketplaces)
+can be represented as an AI Catalog, enabling Claude Code plugins to be
+discovered and indexed alongside other AI artifacts.
 
 ## Overview
 
-The Claude Code Plugins marketplace is defined by a `marketplace.json`
-file that lists available plugins. Each plugin is a directory containing
-a `.claude-plugin/plugin.json` metadata file and optional components:
-MCP server configurations (`.mcp.json`), slash commands (`commands/`),
-agent definitions (`agents/`), and skill definitions (`skills/`).
+A Claude Code plugin marketplace is defined by a `marketplace.json` file
+that lists plugin definitions and tells Claude Code how to install them.
+A plugin source is an installation locator; it is not necessarily a URL
+that retrieves a single artifact document.
 
-```
-marketplace.json                    # Top-level plugin directory
-plugins/
-  example-plugin/
-    .claude-plugin/
-      plugin.json                   # Plugin metadata (name, description, author)
-    .mcp.json                       # MCP server config (optional)
-    commands/                       # Slash commands (optional)
-    agents/                         # Agent definitions (optional)
-    skills/                         # Skill definitions (optional)
-    README.md
-```
+This guide illustrates a projection that catalogs a manifest-backed
+plugin's `.claude-plugin/plugin.json` as its artifact document.
+
+Claude Code's `strict` field controls where component definitions come
+from. With `strict: true` (the default), `plugin.json` is authoritative
+for component definitions, but the marketplace entry may supplement it.
+With `strict: false`, the marketplace entry is the complete definition
+and `plugin.json` is not required. This manifest-backed projection
+applies only when `plugin.json` contains the complete component
+definition. Otherwise, the operator must materialize the effective
+definition or choose another artifact representation.
+
+The guide assumes that the AI Catalog operator is also the marketplace
+operator. When they are different, the catalog operator supplies its own
+`host` information rather than deriving it from marketplace `owner`
+metadata.
+
+The projection is not entirely mechanical. The catalog operator assigns
+globally unique entry identifiers, chooses the artifact representation
+and media type, resolves installation sources to retrievable artifact
+documents, and supplies any verifiable host or publisher identifiers.
 
 ## Conceptual Mapping
 
-| Claude Plugins Marketplace | AI Catalog Equivalent |
-|:---|:---|
-| `marketplace.json` (whole file) | AI Catalog document (top-level) |
-| Marketplace `name` | Catalog `host.displayName` |
-| Marketplace `owner` | Catalog `host` (with `identifier` derived from owner) |
-| `plugins[]` array | Catalog `entries[]` array |
-| Plugin `name` | Entry `identifier` (derived as URN); the plugin manifest carries its own name, so entry `displayName` is omitted |
-| Plugin `description` | Stays in the plugin manifest (which carries its own `description`); entry `description` is omitted to avoid duplicating a value that can drift |
-| Plugin `category` | Entry `tags[]` (first tag) |
-| Plugin `tags` | Entry `tags[]` (merged with category) |
-| Plugin `author` | Entry `publisher` |
-| Plugin `source` (url, git-subdir, or path) | Entry `url` (pointing to the plugin repository) |
-| Plugin `.claude-plugin/plugin.json` | The artifact content (referenced via `url`) |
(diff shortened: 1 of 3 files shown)


darrelmiller (approved):

tadasant (approved):

docs/mappings/claude-code-plugins.md
```diff
@@ -70,9 +66,9 @@ Local path source
 
 ## Marketplace as AI Catalog
 
-The `marketplace.json` from
+The `agent-sdk-dev` plugin from
 [claude-plugins-official](https://github.com/anthropics/claude-plugins-official)
-maps to an AI Catalog where each plugin is an entry:
+maps to an AI Catalog as follows:
```
Tehsmash: I thought this section was to show a conversion from marketplace.json to an ai-catalog.json. The new text doesn't make as much sense I think. Perhaps something like:

> A marketplace.json containing one plugin "agent-sdk-dev" maps to an AI catalog as follows:

This makes me wonder if we can create a tool to do this conversion automatically :/

docs/mappings/claude-code-plugins.md
```diff
@@ -91,48 +87,6 @@ maps to an AI Catalog where each plugin is an entry:
...
-      "metadata": {
-        "homepage": "https://www.adspirer.com"
-      },
-      "trustManifest": {
-        "identity": "urn:claude-plugin:adspirer:ads-agent",
-        "provenance": [
-          {
-            "relation": "publishedFrom",
-            "sourceId": "https://github.com/amekala/adspirer-mcp-plugin",
-            "sourceDigest": "sha1:aa70dbdbbbb843e94a794c10c2b13f5dd66b5e40"
```
Tehsmash, line 113: I understand the reason for not using sourceDigest here as its suppose to be a sha256 content digest of the artifact but having an artifact sourced directly from from a git repo is becoming more and more common.

I'm thinking that we probably need to encode this information in the "version" or "URL" field:

```
"source": {
        "source": "git-subdir",
        "url": "https://github.com/adobe/skills.git",
        "path": "plugins/creative-cloud/adobe-for-creativity",
        "ref": "main",
        "sha": "1307e2c03b9cd20c49872be8cbdfda7ee9aa8c7e"
      },
```

There is a problem with "version" as the plugin.json itself has a version which is not the "sha" and they should probably match. So then I would suggest that we just encode it in the URL itself for example something like:

```
https://github.com/adobe/skills.git/plugins/creative-cloud/adobe-for-creativity@1307e2c03b9cd20c49872be8cbdfda7ee9aa8c7e
```

What I find a little bit odd about this whole conversion is that the "type" is "application/vnd.anthropic.claude-plugin+json" however the given URL's do not resolve to the plugin.json. The "correct" URL for that would be something like:

```
https://raw.githubusercontent.com/adobe/skills/refs/heads/main/plugins/creative-cloud/adobe-for-creativity/.claude-plugin/plugin.json
```

But resolving the plugin.json directly isn't useful as it doesn't come with any information about where to find the plugin "code".

There is also an open question of "how do I know that the URL is a git URL?"

I think in reality we're actually defining here is something like `application/vnd.anthropic.claude-plugin+git` here which implies that you should treat the URL as a git repo which resolves to a plugin folder not a JSON document.

This would then allow you to also have `application/vnd.anthropic.claude-plugin+zip` or whatever which could be a different packaging mechanism.

Tehsmash (changes requested):
The extensions change LGTM. I hadn't realised that we had metadata/extensions on the top level catalog object until I re-read this.

Main issue I have is on the claude code market place conversion stuff, after walking through the examples it seems like we have a bit of a weird situation with the URL and types which needs a bit more discussion IMO.

docs/mappings/claude-code-plugins.md
jonathanhefner: I think the underlying problem is that https://github.com/anthropics/claude-plugins-official/blob/main/.claude-plugin/marketplace.json is not a good basis for an example.  It is large and sprawling, and has already drifted from the excerpt we previously had.

I added a [commit](https://github.com/Agent-Card/ai-catalog/pull/104/commits/a5fcfd6d270792fdeb99aecda70dc613c728cd60) to replace it with a synthetic example.

docs/mappings/claude-code-plugins.md
jonathanhefner: This led down a rabbit hole.  I added a commit to try to address some of the concerns, but if we want to work on this further I would strongly suggest a separate PR.

If desired, I can separate commits https://github.com/Agent-Card/ai-catalog/pull/104/changes/928e4370f7d2251c2d3f1f10d8e5689649f7fbea and https://github.com/Agent-Card/ai-catalog/pull/104/changes/a5fcfd6d270792fdeb99aecda70dc613c728cd60 out of this PR so that it only contains commit https://github.com/Agent-Card/ai-catalog/pull/104/changes/3596257f47b1650685c41e5c5f7e9d0329eaf6dd.

Tehsmash: It really is a rabbit hole, thanks for trying to resolve it!

 https://github.com/Agent-Card/ai-catalog/commit/3596257f47b1650685c41e5c5f7e9d0329eaf6dd is way better already, but my main issue is that pointing to `plugin.json` isn't "useful" for a claude plugin, it'll get you the metadata for the plugin but that doesn't include any reference to the actual plugin code.
 
One option based on https://code.claude.com/docs/en/plugin-marketplaces#plugin-sources is to add different types like:

application/vnd.anthropic.claude-plugin+git
application/vnd.anthropic.claude-plugin+npm
application/vnd.anthropic.claude-plugin+zip

For each of these the catalog entry URL can have a different semantic meaning e.g. +git its a git file distribution, +zip its a zip file. +npm its the NPM package identifier. This is in alignment with how we're doing agent skills.

Alternatively, I think we perhaps need to go up a level and introduce a "plugin card" (sick emoji) which literally just encodes https://code.claude.com/docs/en/plugin-marketplaces#plugin-entries as a separate thing.
You could quite cleanly embed this into AI Catalog something like:

```json
{
   "entries": [
       {
         "type": "application/vnd.anthropic.claude-plugin-entry+json",  <--- note not plugin.json but plugin-entry.json
         "data": {
            {
			      "name": "code-formatter",
			      "source": {
			        "source": "github",
			        "repo": "company/deploy-plugin"
			      },
			      "description": "Automatic code formatting on save",
			      "version": "2.1.0",
			      "author": {
			        "name": "DevTools Team"
			      }
			},
         }
       }  
   ] 
}
```

This pushes all the "sources" stuff out of AI Catalog making it a claude specific thing to deal with. The "plugin entry" schema is closer aligned with something like an MCP server card or A2A card.

mindpower (approved):

PR #105, opened 2026-08-26
Separate artifact identity from registry coordinates

### LINKED ISSUE #102: Design question: how should existing registries map native artifact IDs to urn:air: identifiers?
Hi — I maintain the agent-registry work in [Apicurio Registry](https://github.com/Apicurio/apicurio-registry) (an LF project). We store A2A `AgentCard`s and MCP tool definitions as versioned artifacts, and I'm implementing `/.well-known/ai-catalog.json` as a **projection** over our existing storage so the registry can be crawled by ARD/AI-Catalog clients.

I've hit a design question that I think affects every existing registry adopting the standard, not just us, and I couldn't find it addressed in the spec.

### The tension

The spec **strongly recommends** (and ARD **mandates**, for federation) the identifier form:

```
urn:air:<publisher>:<namespace>:<name>
```

But existing registries already have their own stable, globally-meaningful artifact identity:

- Apicurio keys artifacts by `groupId` / `artifactId` (plus version).
- The MCP Registry keys servers by name.
- OCI registries key by repository path + tag.

When such a registry projects its content into a catalog, there are two plausible mappings and they have different trade-offs:

**Option A — synthesize `urn:air:` from the registry's own coordinates.**
e.g. `urn:air:registry.example.com:<groupId>:<artifactId>`.

- ✅ Stable, unique, no extra user input.
- ❌ The `<publisher>` segment becomes the *registry's* domain, not the *artifact publisher's* domain. For a multi-tenant registry hosting artifacts from many organizations, this conflates the host with the publisher, which seems to cut against the "domain-anchored trust anchor" rationale in Appendix C.

**Option B — require/allow the artifact to declare its own `urn:air:` publisher.**
e.g. the publisher sets it in metadata, and the registry surfaces it verbatim.

- ✅ Preserves the publisher-as-trust-anchor semantics.
- ❌ Breaks the "no extra infrastructure / progressive complexity" goal — now every publisher must mint and manage URNs, and the registry must validate that the declared publisher domain actually matches the serving domain (otherwise trivial namespace squatting, which Appendix C is explicitly trying to prevent).

### What I did (for now)

I went with Option A as the default, with the publisher domain configurable, because it is the only option that works with zero publisher friction. But I'm not confident it's the intended reading, and I suspect other registry implementers are hitting the same fork.

### Ask

Could the spec add guidance (or normative text) on:

1. Whether the `<publisher>` segment, when served from a registry, is meant to be the **registry operator** or the **artifact publisher**?
2. If it's the artifact publisher, what the registry's verification obligation is before it serves a `urn:air:` identifier it didn't mint?
3. Whether a registry-native identifier (e.g. an `https://` URL to the artifact, or an OCI ref) is acceptable as the `identifier` when `urn:air:` can't be honestly constructed — Appendix C reads as "MUST be `urn:air:` for federated systems", which is hard to satisfy truthfully in case (1)-vs-(2) above.

Happy to open a PR with example text once there's a rough consensus on direction. I can also walk through the concrete implementation on a community call if that's easier.

Thanks!

jonathanhefner:
## Summary

This PR proposes a resolution to #102 for existing registries that need to create AI Catalog entries for artifacts without publisher-assigned `urn:air` identifiers.

Publisher-controlled `urn:air` identifiers remain recommended when available, but are no longer required when a registry cannot construct one truthfully.

## Changes Included

- **Added ADR-0021** which documents a stateful preserve-or-mint policy for registries:
   - Reuse an identifier already published for the artifact
   - Otherwise, preserve a publisher-assigned identifier when its use is authorized
   - Otherwise, assign and persist a stable identifier in a namespace controlled by the registry

- **Updated the core specification**:
   - Clarified how consumers handle recognized and unrecognized identifier schemes
   - Added guidance for registry-created entries
   - Prohibited silently replacing an identifier after publication
   - Added an example of preserving native registry coordinates in a namespaced `extensions` value

## Motivation

Requiring every federated entry to use `urn:air:{publisher}:...` leaves existing registries with two problematic choices: present the registry as the artifact publisher, or require publisher enrollment before exposing legacy records.

The proposed preserve-or-mint policy avoids both outcomes. It preserves publisher-controlled identifiers when available while allowing existing records to be exposed without inventing publisher authority. Native registry coordinates remain available for lookup and round trips without being treated as portable artifact identities.

No new core JSON field is required or introduced.

## Scope

This PR intentionally does not change Trust Manifest behavior or define generic aliases and primary-identifier migration. It also leaves distribution-mapping corrections and AIR namespace registration to separate work.

---

**AI usage disclosure:** This PR was drafted with the assistance of AI under my direct supervision and scrutiny.

Closes #102

### DIFF
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -203,9 +203,9 @@ A Catalog Entry object describes a single AI artifact in the catalog.
 It MUST contain the following members:
 
 `identifier`
-: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). Consumers that do not recognize an identifier scheme MUST treat the value as opaque. Identifier syntax alone does not verify publisher identity or establish trust. For open or federated systems, a globally unique absolute URI is RECOMMENDED. An artifact publisher that wants an entry identifier to be preserved when the artifact appears in other catalogs SHOULD use the AI Catalog-specific `urn:air` naming structure with its own domain in the `{publisher}` segment. A catalog incorporating an entry for the first time with a publisher-authorized `urn:air` identifier MUST preserve that identifier exactly.
+: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). Consumers that do not recognize an identifier scheme MUST treat the value as opaque. Identifier syntax alone does not verify publisher identity or establish trust. For open or federated systems, a globally unique absolute URI is RECOMMENDED. An artifact publisher that wants an entry identifier to be preserved when the artifact appears in other catalogs SHOULD use the AI Catalog-specific `urn:air` naming structure with its own domain in the `{publisher}` segment. A catalog operator incorporating an entry for the first time with a publisher-authorized `urn:air` identifier MUST preserve that identifier exactly.
 
-    **AI Catalog Publisher Naming Format:**
+    **`urn:air` Identifier Format:**
     `urn:air:{publisher}:{namespace}:{name}`
 
     - `{publisher}`: The domain name of the organization publishing the artifact (e.g., `example.com`).
@@ -219,7 +219,7 @@ It MUST contain the following members:
 
     For closed or local systems where a different identifier format is used, client implementations are responsible for parsing and processing the custom format as appropriate.
 
-    See [Multi-Version Entries](#multi-version-entries) for uniqueness rules when multiple versions are present, and [Registry Projection](#registry-projection) for identifiers assigned while projecting an existing registry.
+    See [Multi-Version Entries](#multi-version-entries) for uniqueness rules when multiple versions are present, and [Catalog Projection](#catalog-projection) for identifiers assigned while projecting an existing Catalog Entry or source-system record.
 
 `type`
 : A string containing the identifier that specifies the type of the
@@ -435,7 +435,7 @@ When `version` is present, the combination of `identifier` and `version`
 MUST be unique within the catalog. When `version` is absent, `identifier`
 alone MUST be unique. The `identifier` SHOULD be stable across versions.
 Only publisher-authorized `urn:air` identifiers receive a cross-catalog
-preservation requirement, as defined in [Registry Projection](#registry-projection).
+preservation requirement, as defined in [Catalog Projection](#catalog-projection).
 
 Clients that need only the latest version SHOULD sort entries
 sharing the same `identifier` by `version` (when parseable as a semantic
@@ -469,46 +469,44 @@ For example, a catalog listing two versions of the same agent:
 Both entries share the same `identifier` but have distinct `version`
 values, so the combination is unique.
 
-## Registry Projection
+## Catalog Projection
 
 A publisher-authorized `urn:air` identifier is one assigned by the
 artifact publisher or its authorized delegate, with the artifact
 publisher's domain in the `{publisher}` segment.
 
-A registry creating an entry from an existing record or catalog entry
-MUST select its primary identifier by applying these rules in order:
+Catalog projection creates a Catalog Entry from either a source Catalog
+Entry in another AI Catalog or a source record in another system. When
+incorporating a projected entry for the first time, a catalog operator
+MUST select its identifier by applying these rules in order:
 
-1. If the registry previously published an identifier for the artifact,
-   it MUST reuse that identifier, even if another identifier becomes
-   available later.
-2. Otherwise, if the source entry contains a publisher-authorized
-   `urn:air` identifier, it MUST preserve the identifier exactly.
-3. Otherwise, the registry MAY preserve a non-`urn:air` source identifier
+1. If the source contains a publisher-authorized `urn:air` identifier,
+   the operator MUST preserve it exactly.
+2. Otherwise, the operator MAY preserve a non-`urn:air` source identifier
    or replace it. Non-`urn:air` identifiers have no guaranteed portability
    across catalogs.
-4. When assigning a new identifier, a registry that becomes the artifact
+3. When assigning a new identifier, an operator that becomes the artifact
    publisher SHOULD use `urn:air` with its own domain in the `{publisher}`
-   segment. A registry acting as an authorized delegate SHOULD use
+   segment. An operator acting as an authorized delegate SHOULD use
    `urn:air` with the delegating publisher's domain in that segment. An
-   independent registry MUST use a non-`urn:air` identifier under its own
-   control.
+   operator that is neither the artifact publisher nor its authorized
+   delegate MUST use a non-`urn:air` identifier under its own control.
 
-Merely hosting or aggregating an entry does not make a registry the
-artifact publisher.
-A registry MUST NOT infer publisher authorization from the artifact's URL
-or an unsigned `publisher` field.
+Operating a catalog or aggregating an entry does not make its operator the
+artifact publisher. A catalog operator MUST NOT infer publisher
+authorization from the artifact's URL or an unsigned `publisher` field.
 
 Adopting a different primary identifier after publication requires an
 explicit migration mechanism, which this specification does not define.
 
-A registry-assigned identifier identifies the artifact, not a particular
-version or registry record. It MUST remain stable across
-versions, registry-coordinate changes, and retrieval-URL changes, and
-MUST NOT be reassigned to another artifact. The registry assigning the
-identifier, the catalog `host`, and the artifact `publisher` are
-independent roles.
+An identifier assigned by a catalog operator identifies the artifact, not
+a particular version or source record. It MUST remain stable across
+versions, source-coordinate changes, and retrieval-URL changes, and MUST
+NOT be reassigned to another artifact. The catalog operator and artifact
+publisher are distinct roles. The operator is the entity identified by the
+top-level `host` field when that field is present.
 
-A registry MAY retain replaced source identifiers or registry-native
+A catalog operator MAY retain replaced source identifiers or source-system
 coordinates. If retained, they SHOULD be stored in a namespaced entry
 extension. They do not participate in catalog uniqueness or establish
 publisher identity, trust, or equivalence with another identifier.
--- a/specification/ai-catalog.md
+++ b/specification/ai-catalog.md
@@ -492,6 +492,9 @@ MUST select its identifier by applying these rules in order:
    operator that is neither the artifact publisher nor its authorized
    delegate MUST use a non-`urn:air` identifier under its own control.
 
+A projected entry that uses `urn:air` and includes a Trust Manifest remains
+subject to the trust-domain alignment requirements in [Identity](#identity).
+
 Operating a catalog or aggregating an entry does not make its operator the
 artifact publisher. A catalog operator MUST NOT infer publisher
 authorization from the artifact's URL or an unsigned `publisher` field.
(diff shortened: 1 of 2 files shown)


specification/ai-catalog.md
```diff
@@ -203,7 +203,7 @@ A Catalog Entry object describes a single AI artifact in the catalog.
 It MUST contain the following members:
 
 `identifier`
-: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). However, to ensure interoperability, identity uniqueness, and discoverability, the standard `urn:air` naming structure is **HIGHLY RECOMMENDED** and **MUST** be used for open or federated systems.
+: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). Consumers that do not recognize an identifier scheme MUST treat the value as opaque. Identifier syntax alone does not verify publisher identity or establish trust. For open or federated systems, a globally unique absolute URI is RECOMMENDED. The `urn:air` naming structure is RECOMMENDED when the publisher assigns an identifier in a namespace it controls.
```
jonathanhefner: This is the main change to call out.  From:

> the standard `urn:air` naming structure is **HIGHLY RECOMMENDED** and **MUST** be used for open or federated systems

to:

> For open or federated systems, a globally unique absolute URI is RECOMMENDED. The `urn:air` naming structure is RECOMMENDED when the publisher assigns an identifier in a namespace it controls.

The rationale being: federated registries should not invent a `urn:air` identifier for a publisher when the publisher hasn't provided such an identifier for a legacy artifact.  See #102 for more info.

specification/ai-catalog.md
```diff
@@ -203,9 +203,9 @@ A Catalog Entry object describes a single AI artifact in the catalog.
 It MUST contain the following members:
 
 `identifier`
-: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). However, to ensure interoperability, identity uniqueness, and discoverability, the standard `urn:air` naming structure is **HIGHLY RECOMMENDED** and **MUST** be used for open or federated systems.
+: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). Consumers that do not recognize an identifier scheme MUST treat the value as opaque. Identifier syntax alone does not verify publisher identity or establish trust. For open or federated systems, a globally unique absolute URI is RECOMMENDED. An artifact publisher that wants an entry identifier to be preserved when the artifact appears in other catalogs SHOULD use the AI Catalog-specific `urn:air` naming structure with its own domain in the `{publisher}` segment. A catalog incorporating an entry for the first time with a publisher-authorized `urn:air` identifier MUST preserve that identifier exactly.
 
-    **Standard Naming Format:**
+    **AI Catalog Publisher Naming Format:**
```
Tehsmash:
```suggestion
    **AI Catalog Entry Publisher Naming Format:**
```

specification/ai-catalog.md
```diff
@@ -469,6 +469,72 @@ For example, a catalog listing two versions of the same agent:
 Both entries share the same `identifier` but have distinct `version`
 values, so the combination is unique.
 
+## Registry Projection
+
+A publisher-authorized `urn:air` identifier is one assigned by the
+artifact publisher or its authorized delegate, with the artifact
+publisher's domain in the `{publisher}` segment.
+
+A registry creating an entry from an existing record or catalog entry
```
Tehsmash: > existing record

Does this mean a record in a non-ai-catalog registry/market place?

specification/ai-catalog.md
```diff
@@ -469,6 +469,72 @@ For example, a catalog listing two versions of the same agent:
...
+MUST select its primary identifier by applying these rules in order:
+
+1. If the registry previously published an identifier for the artifact,
+   it MUST reuse that identifier, even if another identifier becomes
+   available later.
```
Tehsmash: So if a downstream non-ai-catalog registry renames the artifact / changes the artifacts ID we MUST maintain the ID which we were already using?

Should we see a change of identifier in the downstream catalog as effectively a delete and recreate? And should we basically just do the same on the federated catalog, or do we need to understand that the old identity and new identity are the same somehow?

specification/ai-catalog.md
```diff
@@ -469,6 +469,72 @@ For example, a catalog listing two versions of the same agent:
...
+   `urn:air` identifier, it MUST preserve the identifier exactly.
+3. Otherwise, the registry MAY preserve a non-`urn:air` source identifier
+   or replace it. Non-`urn:air` identifiers have no guaranteed portability
+   across catalogs.
+4. When assigning a new identifier, a registry that becomes the artifact
+   publisher SHOULD use `urn:air` with its own domain in the `{publisher}`
+   segment. A registry acting as an authorized delegate SHOULD use
+   `urn:air` with the delegating publisher's domain in that segment. An
+   independent registry MUST use a non-`urn:air` identifier under its own
+   control.
```
Tehsmash: > An independent registry MUST use a non-`urn:air` identifier under its own control.

Not sure what this last sentence means. What is an independent registry in this case? A registry without a domain?

specification/ai-catalog.md
```diff
@@ -469,6 +469,72 @@ For example, a catalog listing two versions of the same agent:
 Both entries share the same `identifier` but have distinct `version`
 values, so the combination is unique.
 
+## Registry Projection
```
Tehsmash: I think we should re-visit the term "Registry" here, I think it should be something like "Catalog Projection" or something different because if I create a consolidated AI catalog, I'm not running a registry I'm just hosting another AI catalog, and the same as you've described below should apply.

specification/ai-catalog.md
```diff
@@ -469,6 +469,72 @@ For example, a catalog listing two versions of the same agent:
...
+      "registryUri": "https://registry.example",
+      "sourceIdentifier": "foo",
+      "namespace": "payments",
+      "name": "fraud-agent"
+    }
+  }
+}
+```
+
+The extension key is illustrative.
```
Tehsmash, line 537: As we're aware that AI catalog is designed for federation, do you think it makes sense to formalise a native/official extension for federation source / entry origin?

This could then be used to build chain of origin if someone wants to find the normative source of a entry, e.g. resolve the entry origin then check that entries origin until you find an entry with no origin.

jonathanhefner: I pushed a [commit](https://github.com/Agent-Card/ai-catalog/pull/105/commits/dbd96afdd7c0ed93a6cfa58825f7e049ca93909b) to address the other comments, but I think this comment should be addressed in a separate PR if we want to pursue it.

Tehsmash (approved):
Thanks for the update, after this the wording LGTM. 

I agree we can discuss the origin extension as a separate issue to be added later and standardised.

specification/ai-catalog.md
```diff
@@ -469,6 +469,70 @@ For example, a catalog listing two versions of the same agent:
...
+1. If the source contains a publisher-authorized `urn:air` identifier,
+   the operator MUST preserve it exactly.
+2. Otherwise, the operator MAY preserve a non-`urn:air` source identifier
+   or replace it. Non-`urn:air` identifiers have no guaranteed portability
+   across catalogs.
+3. When assigning a new identifier, an operator that becomes the artifact
+   publisher SHOULD use `urn:air` with its own domain in the `{publisher}`
+   segment. An operator acting as an authorized delegate SHOULD use
+   `urn:air` with the delegating publisher's domain in that segment. An
+   operator that is neither the artifact publisher nor its authorized
```
kthota-g, line 492: It would be helpful to note how publisher delegate/operator-assigned identifiers interact with the trust domain validation requirements in [section 5.1](https://ai-catalog.io/spec/#identity). 

Specifically:

> When a Trust Manifest appears within a Catalog Entry, the identity field's trust domain MUST align with the publisher domain in the containing entry's identifier field

When an operator assigns a `urn:air`, it must still satisfy this domain alignment w/ trustManifest.identity.

jonathanhefner: I added a [commit](https://github.com/Agent-Card/ai-catalog/pull/105/changes/5db970bc56a50f6c30397e344f29f70827e89187) with a clarifying sentence:

> A projected entry that uses `urn:air` and includes a Trust Manifest remains subject to the trust-domain alignment requirements in [Identity](https://github.com/jonathanhefner/ai-catalog/blob/5db970bc56a50f6c30397e344f29f70827e89187/specification/ai-catalog.md#identity).

That keeps the normative requirements delegated to the `Identity` section.

Tehsmash (approved):

specification/ai-catalog.md
```diff
@@ -203,9 +203,9 @@ A Catalog Entry object describes a single AI artifact in the catalog.
 It MUST contain the following members:
 
 `identifier`
-: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). However, to ensure interoperability, identity uniqueness, and discoverability, the standard `urn:air` naming structure is **HIGHLY RECOMMENDED** and **MUST** be used for open or federated systems.
+: A string uniquely identifying this artifact. This field is an open text format (e.g., any valid URI or URN is accepted). Consumers that do not recognize an identifier scheme MUST treat the value as opaque. Identifier syntax alone does not verify publisher identity or establish trust. For open or federated systems, a globally unique absolute URI is RECOMMENDED. An artifact publisher that wants an entry identifier to be preserved when the artifact appears in other catalogs SHOULD use the AI Catalog-specific `urn:air` naming structure with its own domain in the `{publisher}` segment. A catalog operator incorporating an entry for the first time with a publisher-authorized `urn:air` identifier MUST preserve that identifier exactly.
```
mindpower, line 206: LGTM. Removing the strict MUST for open/federated catalogs avoids forcing aggregators to mint synthetic URNs for legacy artifacts they don't own. This unblocks Apicurio and existing catalog mirrors while keeping urn:air as the clear best practice

specification/ai-catalog.md
```diff
@@ -469,6 +469,73 @@ For example, a catalog listing two versions of the same agent:
...
+A publisher-authorized `urn:air` identifier is one assigned by the
+artifact publisher or its authorized delegate, with the artifact
+publisher's domain in the `{publisher}` segment.
+
+Catalog projection creates a Catalog Entry from either a source Catalog
+Entry in another AI Catalog or a source record in another system. When
+incorporating a projected entry for the first time, a catalog operator
+MUST select its identifier by applying these rules in order:
+
+1. If the source contains a publisher-authorized `urn:air` identifier,
```
mindpower, line 483: This propagation rule is crucial. Downstream aggregators must not strip or replace an upstream publisher's urn:air with their own host identifier. Preserving the original publisher URN maintains the cryptographic trust chain through federation.

mindpower (approved):
LGTM

muscariello (approved):
@jonathanhefner  

One interaction with #108  worth a note: 

ADR-0021 allows two operators to mint different identifiers for the same artifact absent a publisher-authorized urn:air.  #108 requires subject.identifier to exactly equal entry.identifier, so a signed manifest can't survive being projected into a catalog that mints a different identifier for it. Might be worth documenting what happens to an existing signature in that case.

darrelmiller (approved):

darrelmiller:
@tadasant  @dsp-ant  Can we get one more TSC vote for this PR?

PR #108, opened 2026-09-06
Bind signed Trust Manifests to release coordinates

jonathanhefner:
## Summary

A signed Trust Manifest currently binds its claims to an artifact representation through `subject.type`, `subject.digest`, and optionally `subject.url`. The containing Catalog Entry’s `identifier` and `version` remain outside that signature, even though they determine the artifact’s logical name and participate in catalog version selection.

A catalog writer can therefore retain a valid signed manifest and the exact artifact bytes while assigning them to another logical artifact or relabeling an old release with a higher version. In the latter case, latest-version or minimum-version selection can return vulnerable code while every existing per-entry signature, type, URL, and digest check succeeds.

This PR requires every Subject to identify the artifact it describes. For a Subject on a Catalog Entry, `subject.identifier` must match `entry.identifier`. For a signed entry Trust Manifest, `entry.version` and `subject.version` must either both be omitted or both be present and exactly equal. This gives the signed version one unambiguous role as an authenticated restatement of the entry’s release coordinate.

ADR-0025 records the decision and its rationale. The specification, threat model, CDDL, examples, mapping guidance, and author documentation are updated accordingly.

## Rationale

A top-level catalog signature authenticates the catalog signer’s snapshot, including its identifiers and versions. It does not necessarily establish that the Trust Manifest signer assigned those release coordinates, because the catalog operator and artifact publisher may be different principals. Conversely, an entry Trust Manifest does not protect collection-wide structure. The two signatures make different statements and provide complementary guarantees.

Keeping `subject.url` optional preserves mirror and relocation support. The signed logical identifier, applicable version, media type, and digest are sufficient to bind the claims to an artifact release and its exact representation without binding that release to one location.

## Scope

This PR does not redefine `TrustManifest.identity` or specify signing-key selection for `did:web`, SPIFFE, or other identity mechanisms. It also does not define what the `subject` of a signed Host Trust Manifest represents. The current specification requires such a manifest to carry a subject without defining its Host-specific meaning; that pre-existing discrepancy requires a separate decision and should be reviewed in concert with this proposal.

---


specification/ai-catalog.md
```diff
@@ -660,15 +662,22 @@ provenance:
...
+describes, closing substitution gaps in which an attacker who controls the
+catalog document leaves a validly-signed Trust Manifest in place but repoints
+the entry, assigns the manifest to another logical artifact, or relabels the
+artifact as another version. Because the Subject is part of the signed
+payload, the release coordinates, artifact reference when present, and content
+digest cannot be changed without invalidating the signature.
 
 A Subject object MUST contain:
 
+`identifier`
```
mindpower, line 675: +1. This is a great call to include the identifier here for the trust binding.

specification/ai-catalog.md
```diff
@@ -1714,10 +1746,21 @@ TrustManifest = {
...
   ? privacyPolicyUrl: text,
   ? termsOfServiceUrl: text,
+  ? subject: Subject,
+  ? issuedAt: tdate,
+  ? expiresAt: tdate,
   ? signature: text,
   ? extensions: { * text => any }
 }
 
+Subject = {
```
mindpower, line 1756: Looks good. This is the right follow up to address the security concerns raised in #47

mindpower (approved):
LGTM

muscariello (approved):
Mechanism and examples check out end to end. 

@jonathanhefner  one gap before merge: 

subject.version is only required to match when entry.version is present, so an entry can just drop its version field to escape the comparison entirely. 

The same relabeling attack this PR closes, via deletion instead of forgery. I'd suggest making the binding symmetric: if either side has a version, both must, and must match.

darrelmiller (approved):

darrelmiller:
@tadasant  @dsp-ant   Could we get one more TSC vote to get this merged? Thanks.

PR #109, opened 2026-09-06
Remove Host Trust Manifests from v1

jonathanhefner:
## Summary

The specification currently permits a Trust Manifest on either a Catalog Entry or a Host Info object, but defines the manifest and its verification in artifact-specific terms. In particular, a signed Trust Manifest requires a `subject` whose media type and digest are verified against a containing Catalog Entry. Host Info has no corresponding artifact, media type, URL or data value, or bytes for a consumer to verify.

This leaves no defined way to construct or verify the signed Host Trust Manifest permitted at Level 3. Signing only the nested manifest would protect its internal fields, but would not approve the surrounding Host Info, its placement in a catalog, or the catalog snapshot. The specification also defines no Host-specific claim vocabulary or independent consumption workflow that would justify that separate signature and lifecycle.

This PR removes `host.trustManifest` and defines Trust Manifests exclusively as artifact-oriented companions to Catalog Entries. ADR-0026 records the decision and deferred alternatives. The normative prose, CDDL, data-model diagram, conformance rules, verification guidance, threat model, and author documentation are updated accordingly.

## Rationale

An Entry Trust Manifest has a concrete security purpose: its signed `subject` binds publisher claims to an artifact that a consumer can retrieve and verify. Host Info instead identifies the operator of the catalog. Its integrity and association with a particular snapshot belong at catalog scope.

The optional top-level catalog signature already covers Host Info as part of the catalog document. Once the catalog signer is independently authorized, that signature has the appropriate byte scope for authenticating the operator metadata together with the catalog it describes. A nested Host signature is neither a substitute for that catalog signature nor evidence that the Host approved the surrounding catalog.

Independently signed audit reports and organizational credentials can retain their native proof formats. A future profile may define how a catalog carries and evaluates such operator evidence after a concrete producer, consumer, verification procedure, and trust decision are established. Until then, implementations can incubate operator-specific evidence through `AICatalog.extensions` without committing the core v1 model to undefined semantics.

## Scope

This PR does not change Entry Trust Manifest contents or verification. The separate release-coordinate binding proposal remains responsible for binding signed Entry Trust Manifests to logical artifact identifiers and versions.

This PR also does not repair the existing omission of the top-level `signature` from the normative CDDL or define how a catalog signature selects and authorizes a signer or verification key. The threat model now records catalog-operator authentication as a residual risk; completing the catalog-signature profile requires a separate decision.

---


mindpower (approved):

mindpower (approved):

muscariello (approved):
LGTM

Clean removal.

No leftover Host Trust Manifest references anywhere in the repo, and the threat-model update honestly documents the new residual risk (S3 host-spoofing loses its mitigation) rather than hiding it. 

Non-blocking nit: dropping list item 3 in "Verifying Host Identity" also dropped the blank line before the next heading — cosmetic, builds fine.

darrelmiller (approved):

darrelmiller:
@tadasant   @dsp-ant  Can we get another TSC vote to get this change over the line? Thanks.

PR #117, opened 2026-09-13
Group trust metadata by contributor and generalize signatures

jonathanhefner:
**This draft is based on top of #108, #109, and #110, which have all been approved.  It will be rebased onto `main` once those PRs are merged.  Until then, the [last commit](https://github.com/Agent-Card/ai-catalog/pull/117/changes/0fa896dbc094b4c596ca598db48c4c4ce7ad39f4) in this PR contains the relevant changes -- preceding commits belong to the prerequisite PRs.**

---

## Summary

Allow publishers, assessors, and federated registries to contribute trust metadata about the same artifact, with independent signatures over the fields each endorses.

| Area | Current design | Proposed design |
| --- | --- | --- |
| Trust metadata | One `entry.trustManifest` | `entry.trustManifests[identity]`, each retaining its attestation and provenance arrays |
| Signatures | Entry signature covers the whole Trust Manifest; separate root signature | Separate `signatures` arrays on entries and the catalog root, selecting fields through arrays of keys |
| Artifact binding | Release fields repeated in `entry.trustManifest.subject` | Sign the entry’s release fields and `entry.digest` directly |
| Policy URLs | Inside the signed Trust Manifest | Shared entry fields; signing coverage follows the selected paths |

## Motivation

- **Independent contributions.** A publisher can sign its source provenance while an assessor signs its audit evidence for the same release. If each selects its own Trust Manifest and the shared release fields, the assessor can update its evidence without invalidating the publisher’s signature. A registry can add its own contribution alongside both.
- **Flexible coverage.** Signers can endorse entry extensions or other relevant fields. Catalog operators can sign Host Info through root signatures or endorse the complete catalog snapshot.
- **Reuse existing standards.** Detached JWS and JCS authenticate selected paths, values, and timestamps. The existing `did:web` rules authenticate independent contributors while preserving the additional checks for publisher authority.

## Rationale

[ADR-0028](https://github.com/jonathanhefner/ai-catalog/blob/0fa896dbc094b4c596ca598db48c4c4ce7ad39f4/adr/0028-identity-keyed-trust-manifests.md) explains the construction and tradeoffs.

The schema, guides, mappings, examples, and threat model are updated together.

## Notes

- **Dependencies:** [#108](https://github.com/Agent-Card/ai-catalog/pull/108), [#109](https://github.com/Agent-Card/ai-catalog/pull/109), and [#110](https://github.com/Agent-Card/ai-catalog/pull/110).
- **Validation:** Specification and strict documentation builds, example checks, and local signing tests.


adr/0028-identity-keyed-trust-manifests.md
```diff
@@ -44,7 +44,7 @@ The existing fields map to the proposed structure as follows:
 
 | Existing location | Proposed location |
 | --- | --- |
-| `trustManifest.identity` | Key in `entry.trustManifests` |
+| `trustManifest.identity` | Contributor key in `entry.trustManifests`; endorsement identity in `entry.signatures[].signer` |
```
muscariello, line 47: One gap: a future signer profile could accept the same kind of signer value as did:web. 

A did:web:acme.com:... identity and the plain URL https://acme.com/... refer to the same underlying resource. So imagine a second profile gets added later — say, a plain JWKS-over-HTTPS profile that just fetches a key list from a URL, no extra authorization check. If a verifier can't tell the two profiles apart from the signer value alone, it might apply the JWKS profile's rules to a signer that was actually meant to go through did:web. That matters because the JWKS profile would accept any key found at that URL, while did:web only accepts a key the DID document specifically lists under assertionMethod. Applying the JWKS profile by mistake means skipping that check entirely, not just using a different lookup mechanism.

To fix that,  add an explicit profile field to the Signature object, signed the same way signer is today. Same paths/issuedAt/jws shape either way — only signer and profile differ:

ex:

{
  "signer": "did:web:acme-corp.com:agent:finance",
  "profile": "did:web-v1",
  "paths": [["identifier"], ["type"], ["digest"], ["trustManifests", "did:web:acme-corp.com:agent:finance"]],
  "issuedAt": "2026-09-13T00:00:00Z",
  "jws": "..."
}

{
  "signer": "https://acme.com/certs/jwks.json",
  "profile": "https-jwks-v1",
  "paths": [["identifier"], ["type"], ["digest"], ["trustManifests", "https://acme.com/certs/jwks.json"]],
  "issuedAt": "2026-09-13T00:00:00Z",
  "jws": "..."
}

A verifier picks the profile by reading profile directly, not by guessing from what signer looks like — and since it's part of the signed payload, it can't be changed after the fact. Small addition today, much harder to bolt on safely once a second profile actually exists.

muscariello (commented):
See inline

muscariello:
I would like to share how I would use this to also include SPIRE/SPIFFE not as workload identity manager but as a signing identity:

spiffe-x509-v1 profile could work like this:

signer = the SPIFFE ID (spiffe://acme.com/ns/finance/sa/finance-agent).

Verification fetches the trust domain's bundle endpoint (live, same as did:web resolves a DID document live) to get the current root/intermediate CAs — this part rotates rarely, same cadence as a DID document.


The actual short-lived leaf SVID that did the signing doesn't need re-resolving at all — it rides along in the JWS's own protected header via the standard x5c parameter (RFC 7515's mechanism for embedding a certificate chain directly in a JWS), so no new Signature-object field is needed.


Verification checks: 
the x5c chain is cryptographically valid and terminates at one of the currently fetched trust-domain roots, the leaf cert's SAN contains the claimed SPIFFE ID, and issuedAt falls inside the leaf cert's own validity window — 
"was this valid then," not "is it valid now." That's the piece that actually accounts for SVIDs being short-lived: you don't need the leaf to still be valid at verification time, you need proof it was valid at signing time, exactly like a timestamped code-signing certificate that outlives its own cert's expiry.
It's equivalent to sigstore.

muscariello:
did:web-v1
{
  "signer": "did:web:acme-corp.com:agent:finance",
  "profile": "did:web-v1",
  "paths": [["identifier"], ["type"], ["digest"], ["trustManifests", "did:web:acme-corp.com:agent:finance"]],
  "issuedAt": "2026-09-13T00:00:00Z",
  "jws": "..."
}

https-jwks-v1
{
  "signer": "https://acme.com/certs/jwks.json",
  "profile": "https-jwks-v1",
  "paths": [["identifier"], ["type"], ["digest"], ["trustManifests", "https://acme.com/certs/jwks.json"]],
  "issuedAt": "2026-09-13T00:00:00Z",
  "jws": "..."
}

spiffe-x509-v1
{
  "signer": "spiffe://acme.com/ns/finance/sa/finance-agent",
  "profile": "spiffe-x509-v1",
  "paths": [["identifier"], ["type"], ["digest"], ["trustManifests", "spiffe://acme.com/ns/finance/sa/finance-agent"]],
  "issuedAt": "2026-09-13T00:00:00Z",
  "jws": "..."
}

jonathanhefner:
@muscariello Thank you for the review!

I've pushed https://github.com/Agent-Card/ai-catalog/pull/117/changes/51df07b58e831f71ab6047672d6482ef93f5015f, which adds a `profile` field to signature objects.

specification/examples/ai-catalog.json
```diff
@@ -10,45 +10,47 @@
...
+            {
+              "type": "SOC2-Type2",
+              "uri": "https://trust.acme-corp.com/reports/soc2-2026.pdf",
+              "digest": "sha256:a1b2c3d4e5f67890abcdef1234567890abcdef1234567890abcdef1234567890",
+              "size": 245760
+            }
+          ]
+        }
+      },
+      "signatures": [
```
mindpower, line 33: The idea of grouping claims by contributor (trustManifests) is great for downstream evaluations. However, moving signatures to a top-level entry.signatures: [...] array with 2D JSON path selectors introduces substantial verification fragility, custom path-walking parsers across all SDKs, and breaks the 'self-contained envelope' principle. 

Keeping signature and subject inside each trustManifest keeps manifests transport-portable, and avoids custom JSON path-walking engines."

jonathanhefner: Path selectors allow targeted signing.  Without that we would need to:

- Duplicate entry data in the `subject` field
- Choose exactly what is allowed in the `subject` -- e.g., is `extensions` allowed?
- Remember to allow new fields in the `subject` -- e.g., `version` was omitted before #108

> custom JSON path-walking engines

I feel like that sounds a lot more complex than it actually is. 😅  Here is an example implementation:

```typescript
function buildPayload(paths: string[][], entry: JsonObject) {
  return paths
    .toSorted(comparePaths)
    .map(path => [path, extractValue(path, entry)] as const);
}

function comparePaths(a: string[], b: string[]): number {
  for (let i = 0; i < Math.min(a.length, b.length); i++) {
    if (a[i] < b[i]) return -1;
    if (a[i] > b[i]) return 1;
  }
  return a.length - b.length;
}

function extractValue(path: string[], entry: JsonObject): JsonValue {
  if (path[0] === "signatures") {
    throw new Error("Selecting top-level signatures is forbidden");
  }

  let value: JsonValue = entry;

  for (const key of path) {
    if (!isJsonObject(value) || !Object.hasOwn(value, key)) {
      throw new Error(`Cannot resolve path: ${JSON.stringify(path)}`);
    }
    value = value[key];
  }

  return value;
}

function isJsonObject(value: JsonValue): value is JsonObject {
  return value !== null && typeof value === "object" && !Array.isArray(value);
}
```

If we want to simplify even further, we can drop the sorting requirement and allow `signatures` to be included in a path (though no one should want to).

If you're strongly opposed to selector paths, though, I can explore less generic approaches.

jonathanhefner: An example of a less general approach would be to replace `paths` with:

- implied `entry.identifier`, `entry.type`, `entry.version` if present, `entry.digest` if present
- optional `signatures[].additionalSubjectFields` array of entry fields to include in an ephemeral (generated) subject
- optional `signatures[].subjectExtensions` array of extension keys to include in an ephemeral subject
- optional `signatures[].subjectTrustManifests` array of Trust Manifest contributor identities to include in an ephemeral subject

Of course, names can be tweaked -- e.g., `signatures[].additionalEntryFields` or `signatures[].subjectDefinition.entryFields` or whatever.

(By the way, even if we decide to keep `paths`, we can make the inclusion of `entry.identifier` et al implied.)

I still favor keeping `signatures` as a peer of `trustManifests` though.  Putting `signatures` inside a Trust Manifest makes it a bit more awkward to generate the payload to sign (because `signatures` would have to be removed from each Trust Manifest before signing).

muscariello: I agree with @jonathanhefner that getting signatures in the payload is ugly.

jonathanhefner: If we really want all trust-related data contained in a single field, I would recommend to:

- rename `trustManifests` to `trustContributions` or just `contributions`
- make a _newly defined_ Trust Manifest object (`entry.trustManifest`) that would contain `contributions` and `signatures`
- possibly move `privacyPolicyUrl` and `termsOfServiceUrl` from directly under `entry` back into individual `contributions` objects
  - this would partly be based on who we expect to contribute these values, but also partly based on how those fields are referenced to build an ephemeral subject
- keep `digest` directly under `entry`

Altogether, an **entry** could look like:

```json
{
  "identifier": "urn:air:acme.com:agent:finance-a2a",
  "type": "application/a2a-agent-card+json",
  "version": "2.1.0",
  "digest": "sha256:...",
  "url": "..."
  "publisher": {
    "identifier": "did:web:acme.com",
    "displayName": "Acme Financial Corp"
  },
  "extensions": {
    "https://ai-catalog.org/extensions/metadata": {
      "foo": "bar",
    }
  },
  "trustManifest": {
    "contributions": {
      "did:web:acme.com": {
        "privacyPolicyUrl": "https://acme.com/legal/privacy",
        "termsOfServiceUrl": "https://acme.com/legal/terms",
        "attestations": [
          {
            "type": "SOC2-Type2",
            "uri": "https://trust.acme.com/reports/soc2.pdf",
            "digest": "sha256:..."
          }
        ]
      }
    },
    "signatures": [
      {
        "signer": "did:web:acme.com",
        "profile": "did-web-v1",
        "subjectAdditionalFields": ["url"],
        "subjectExtensions": ["https://ai-catalog.org/extensions/metadata"],
        "subjectTrustContributions": ["did:web:acme.com"],
        "issuedAt": "2026-03-20T14:00:00Z",
        "jws": "..."
      }
    ]
  }
}
```

mindpower: Thanks @jonathanhefner for the detailed explanation and the prototype code, and @muscariello for commenting on the parsing challenges of enveloped signatures.

I agree that enveloped signatures (where a signature sits inside the signed document and must be stripped before verification) introduce parsing fragility and canonicalization edge cases.

However, introducing a custom, path-selector DSL (`paths: string[][]`) to selectively sign document fragments introduces significant new security and interoperability challenges:

1. **Selective Signing / Truncation Vulnerability:**
   When a signature only commits to specific selected paths (e.g. `[["identifier"], ["version"]]`), any untrusted registry or MITM can inject arbitrary unselected fields (e.g. malicious `endpoints` or `permissions`) into the entry, and the signature will still validate cleanly. The verifier has no cryptographic guarantee over what was deliberately omitted versus what was maliciously injected.
2. **Cross-Language Canonicalization & Interop (RFC 8785):**
   Converting an in-memory tuple list `[[path, value], ...]` into deterministic, byte-for-byte signing bytes requires strict canonicalization across all SDKs (Python, Go, Rust, TypeScript). Array indexing, missing keys, and path overlap handling will create spec divergence and parsing vulnerabilities.
3. **The Role of `subject` (Cryptographic Digest vs. Data Duplication):**
   Jonathan raised valid concerns about needing to duplicate entry fields in `subject` or constantly updating the `subject` schema when new fields (like `version` or `extensions`) are added. 
   
   In established supply-chain standards (such as **in-toto Statements** and **SLSA**), the `subject` is **never** a duplicate copy of entry fields. It is simply an immutable cryptographic digest:
   ```json
   "subject": [
     {
       "name": "entry:identifier",
       "digest": { "sha256": "<hash-of-canonical-entry>" }
     }
   ]

In this case, it's y binding via content digest (sha256 of RFC 8785 canonical JSON). so the verifier never duplicate entry data.

mindpower: Regarding the latest splitting trustManfiest into **trustContribution** and **signatures**:

The good:
1. Moving away from arbitrary 2D JSON paths (`paths: string[][]`).
2. Establishing `identifier`, `type`, `version`, and `digest` as non-negotiable core fields.
3. Consolidating everything under `entry.trustManifest` with `contributions` and `signatures` (Comment #2). This brings much-needed structural cohesion to the catalog entry.

However, the idea of an **ephemeral (dynamically synthesized) subject** via `subjectAdditionalFields`, `subjectExtensions`, and `subjectTrustContributions` still introduces serious security and interoperability issues: (similar to the 2D json path)

1. The Selective Signing / Truncation Vulnerability (Confused Deputy)
If a signer only binds to an ephemeral subset of fields:
- What prevents an untrusted intermediary or rogue registry from modifying unselected fields (e.g. altering `publisher.displayName`, changing `url`, or injecting an unauthorized extension)?
- The signature over the ephemeral subject would still verify as 100% valid! A consumer checking the signature would see a valid signature from the publisher and falsely assume the entire entry is authentic.
- 
2. Verifier Policy and Cross-Language Drift
In this model, every signature in `signatures` can have a different recipe for what it chose to include or omit. Verifiers in Python, Go, Rust, and TypeScript would not only have to implement complex plucking-and-reassembly logic, but consuming applications would have to inspect each signature's recipe to determine if critical fields were left unsigned.

mindpower: Basically, with path selectors and ephemeral subjects, we are trying to re-invent an ad-hoc JSON path-walking engine.

As discussed earlier, while the TypeScript prototype snippet looks deceptively simple, from a security and standards engineering standpoint, it introduces critical vulnerabilities.

The latest code example is largely moving in the right direction, except that instead of manual field selection, we should sign the whole entry and the whole contribution respectively. 

It would look like this:

```json
{
  "identifier": "urn:air:acme.com:agent:finance-a2a",
  "version": "2.1.0",
  "digest": "sha256:7f83b1657ff...",
  "url": "https://agents.acme.com/finance",
  "publisher": { "identifier": "did:web:acme.com" },
  "extensions": { "foo": "bar" },

  "signature": {
    "signer": "did:web:acme.com",
    "jws": "<publisher-signature-over-entry>"
  },

  "trustManifests": [
    {
      "contributor": "did:web:acme.com",
      "subject": { "digest": "sha256:7f83b1657ff..." },
      "privacyPolicyUrl": "https://acme.com/legal/privacy",
      "attestations": [ ... ],
      "signature": {
        "signer": "did:web:acme.com",
        "jws": "<signature-over-acme-manifest>"
      }
    },
    {
      "contributor": "did:web:cisco.com",
      "subject": { "digest": "sha256:7f83b1657ff..." },
      "attestations": [
        {
          "type": "SecurityPostureAudit",
          "score": "A+"
        }
      ],
      "signature": {
        "signer": "did:web:cisco.com",
        "jws": "<signature-over-cisco-manifest>"
      }
    }
  ]
}

Comment (jonathanhefner) [reply specification/examples/ai-catalog.json]: > 1\. **Selective Signing / Truncation Vulnerability:**
>    When a signature only commits to specific selected paths (e.g. `[["identifier"], ["version"]]`), any untrusted registry or MITM can inject arbitrary unselected fields (e.g. malicious `endpoints` or `permissions`) into the entry, and the signature will still validate cleanly. The verifier has no cryptographic guarantee over what was deliberately omitted versus what was maliciously injected.

> 1\. The Selective Signing / Truncation Vulnerability (Confused Deputy)
>    If a signer only binds to an ephemeral subset of fields:
> 
> * What prevents an untrusted intermediary or rogue registry from modifying unselected fields (e.g. altering `publisher.displayName`, changing `url`, or injecting an unauthorized extension)?
> * The signature over the ephemeral subject would still verify as 100% valid! A consumer checking the signature would see a valid signature from the publisher and falsely assume the entire entry is authentic.

That also applies to what we have now, no?  Isn't that just the nature of (intentionally) signing a subset of values?

---

> 2\. **Cross-Language Canonicalization & Interop (RFC 8785):**
>     Converting an in-memory tuple list `[[path, value], ...]` into deterministic, byte-for-byte signing bytes requires strict canonicalization across all SDKs (Python, Go, Rust, TypeScript). Array indexing, missing keys, and path overlap handling will create spec divergence and parsing vulnerabilities.

I'm not sure I understand the point this is trying to make.

The same applies to JSON objects (perhaps even moreso).  We also already have arrays that we are signing, such as `attestations`.

---

> 3\. **The Role of `subject` (Cryptographic Digest vs. Data Duplication):**
>     Jonathan raised valid concerns about needing to duplicate entry fields in `subject` or constantly updating the `subject` schema when new fields (like `version` or `extensions`) are added.
>     In established supply-chain standards (such as **in-toto Statements** and **SLSA**), the `subject` is **never** a duplicate copy of entry fields. It is simply an immutable cryptographic digest

Producing a an immutable cryptographic digest requires an ephemeral structure to hash...

If we're going define and materialize a structure such that it could be hashed, why not just include the structure directly in the signing payload?

In any case, _what is the definition of that structure_?

---

> 2\. Verifier Policy and Cross-Language Drift
>     In this model, every signature in `signatures` can have a different recipe for what it chose to include or omit. Verifiers in Python, Go, Rust, and TypeScript would not only have to implement complex plucking-and-reassembly logic, but consuming applications would have to inspect each signature's recipe to determine if critical fields were left unsigned.

Yes, that is by design.  Different signers may sign different things -- e.g., signing the URL vs not, signing different trust manifests, signing different extensions.

---

> The latest code example is largely moving in the right direction, except that instead of manual field selection, we should sign the whole entry and the whole contribution respectively.
> 
> It would look like this:

@mindpower Signing the whole entry means extensions cannot be added and `url` cannot be unsigned / changed for mirrors.

Therefore, I think you're saying "sign a slice of the entry".  In which case, _what is the slice_ and _do we allow variations from it_ (such as additional fields)?

Comment (mindpower) [reply specification/examples/ai-catalog.json]: > That also applies to what we have now, no? Isn't that just the nature of (intentionally) signing a subset of values?



That's different. The publisher simply signs the entry data (the entry object excluding `signature` and `trustManifests`). When a publisher publishes an entry, they stand behind that complete definition. Allowing arbitrary slicing is what creates confused-deputy risks.

> Different signers may sign different things

Agreed, but they should sign *different self-contained objects* (Acme signs the entry; Cisco signs Cisco’s contribution), not arbitrary slices. Also `url` should not be mutable. If an intermediary or mirror can rewrite `url` without invalidating the publisher's signature, an attacker can redirect users to a malicious endpoint while keeping Acme’s valid signature.

Comment (jonathanhefner) [reply specification/examples/ai-catalog.json]: > That's different. The publisher simply signs the entry data (the entry object excluding `signature` and `trustManifests`). When a publisher publishes an entry, they stand behind that complete definition. Allowing arbitrary slicing is what creates confused-deputy risks.

To reiterate my clarification from the meeting: my follow-up proposal in https://github.com/Agent-Card/ai-catalog/pull/117#discussion_r4063033958 did not allow completely arbitrary slicing.  All signatures would include the same core entry fields that the publisher signs.

> Agreed, but they should sign _different self-contained objects_ (Acme signs the entry; Cisco signs Cisco’s contribution), not arbitrary slices. Also `url` should not be mutable. If an intermediary or mirror can rewrite `url` without invalidating the publisher's signature, an attacker can redirect users to a malicious endpoint while keeping Acme’s valid signature.

Also reiterating from the meeting: `entry.digest` protects the artifact while allowing mirror URLs.  The spec currently allows that even without this PR.  If we want to disallow mirror URLs, it would be an additional behavioral change on top of what this PR is proposing.

However, I understand your preference for self-contained objects.  During the meeting, @muscariello also made the point that when there are multiple signers, they often sign the exact same (immutable) object.

Taking all of the above into account, I the following structure is closer to what you're asking for:

```jsonc
{
  "identifier": "urn:air:acme.com:agent:finance-a2a",
  "type": "application/a2a-agent-card+json",
  "version": "2.1.0",
  "digest": "sha256:...",
  "url": "...",
  "publisher": {
    "identifier": "did:web:acme.com",
    "displayName": "Acme Financial Corp"
  },
  "extensions": {
    "https://ai-catalog.org/extensions/metadata": {
      "foo": "bar",
    }
  },

  "trustManifests": [
    {
      "contributor": "did:web:acme.com",
      "privacyPolicyUrl": "https://acme.com/legal/privacy",
      "termsOfServiceUrl": "https://acme.com/legal/terms",
      "attestations": [
        {
          "type": "SOC2-Type2",
          "uri": "https://trust.acme.com/reports/soc2.pdf",
          "digest": "sha256:..."
        }
      ],

      // `canonicalUrl` would be semantically similar to `entry.url` but not
      // required to exactly match. It would merely communicate the artifact
      // URL that the Trust Manifest contributor endorses.
      //
      // **THIS FIELD COULD BE DEFINED IN V1.1 OF THE SPEC**
      "canonicalUrl": "...",

      // `supportedExtensions` lists the extensions (by key) that the Trust
      // Manifest contributor supports. If those extensions are present in the
      // entry, then their values MUST be included in the signature payload -- i.e.,
      // anyone signing this Trust Manifest must also sign exactly those extensions.
      //
      // **THIS FIELD COULD BE DEFINED IN V1.1 OF THE SPEC**
      "supportedExtensions": ["https://ai-catalog.org/extensions/metadata"],

      "signatures": [
        {
          "signer": "did:web:acme.com",
          "profile": "did-web-v1",
          "issuedAt": "2026-03-20T14:00:00Z",

          // The actual JWS signature covers:
          // - `entry.identifier`, `entry.type`, `entry.digest`, and `entry.version` if present
          // - the present `entry.extensions` values listed in `supportedExtensions`
          // - the containing Trust Manifest object minus the `signatures` field
          // - this Signature object minus the `jws` field
          "jws": "..."
        }
      ]
    }
  ]
}
```

It does put `signatures` back inside the Trust Manifest, which I think is a bit awkward.  If we don't like that, then we can make `signatures` a peer of `trustManifests` (or `trustManifest.contributions`), and each signature object could contain a single Trust Manifest key.

Also note that if we want to leave the `canonicalUrl` and `supportedExtensions` fields out of v1 of the spec, then we can do so without painting ourselves into a corner (_assuming we think they are at least directionally correct_).

